Issue #33, opened 2016-11-16
Create release.py


lboxell:
See https://github.com/gslab-econ/template/issues/19#issuecomment-260826983. Should either live in gslab_python or at the root of the template and its helpers live in gslab_python. 

See the next two comments also.

stanfordquan:
With #36 pulled in, is this issue completed?

M-R-Sullivan:
@stanfordquan - No, as I still need to look into an error that @arosenbe encountered  in [this issue's /template/ pull request](https://github.com/gslab-econ/template/pull/21). If the completion of this issue is a prerequisite for a priority of yours, I can focus on resolving this error in coming work hours.

stanfordquan:
Not at all, don't worry. I looked over the commit history and looks like unit tests are done and dusted, so I don't need to work on that! Thanks.

M-R-Sullivan:
This issue has passed peer review at #36 and https://github.com/gslab-econ/template/pull/21. So I am closing it now.

**Summary**
- I released GSLab Tools 3.0.0, which adds functionality for making releases from SCons/GitHub repositories to GitHub and to Google Drive to `gslab_scons`. This functionality is described in [gslab_scons/README.md](https://github.com/gslab-econ/gslab_python/blob/master/gslab_scons/README.md).
- The `release.py` script `gslab_scons` checks that versioned files are not overly large and that the repository from which the user is making a release is up-to-date (i.e. that SCons has built all of the repository's targets and that changes have been committed with git). 
- I revised code in the [large template](https://github.com/gslab-econ/template/tree/large_template) to be compatible with `gslab_scons`'s new release functionality.

Issue #28, opened 2016-11-11
gslab_scons restructure


lboxell:
Follow package organization established in #24. 

* For check_code_extension(), why not just pass it the extension directly rather than the software name; so we could say in build.py:

```
check_code_extension(source_file, '.do')
```

and in misc.py

```
def check_code_extension(source_file, ext):
    source_file = str.lower(source_file)
    if not source_file.endswith(ext):
        raise BadExtensionError('First argument, ' + source_file + ', must be a ' + ext + ' file')
    return None
```

* Having `target_file` in the code makes it look like we require there to be a single target file. Really, we're only using this to pick a directory for the log file. So I might drop the lines starting with `target_file` and `target_dir` and instead say

```
log_dir = os.path.dirname( str(target[0]) )
``` 

* Should we have all builder's return an informative error message for the case where the executable is not found / installed (so, e.g., if I call the Lyx builder and don't have lyx correctly on the path I get a nice error explaining what happened and suggesting that I make sure it's installed and on the path)? This is a very common case when people try to run our code, and it would be nice to handle it explicitly.

* In build tables we can replace `source[1:len(source)]` with `source[1:]`

* I think we should restructure build_stata. I would suggest we define two custom functions get_stata_executable() and get_stata_options() so that the main builder ends with

```
exec = get_stata_executable('user_flavor')
options = get_stata_options()

os.system('%s %s' % (exec, options))
```

and is otherwise the same as the other builders. Then we could structure get_stata_executable() as roughly:

```
def get_stata_executable(user_flavor):
    exec_list = ['stata-mp', 'stata-mp.exe', 'stata-mp-64.exe', 'stata-se', ...etc...]

    if user_flavor:
        exec = user_flavor
    elif platform == 'win32' & os.environ['STATAEXE']:
        exec = os.environ['STATAEXE']
    else: 
        for e in exec_list:
            if is_in_path(e):
                exec = e
    return exec
```

* start_log should include:

```
if not (mode in ['develop', 'cache', 'release']):
	print("Error: %s is not a defined mode" % mode)
	sys.exit()

if mode == 'release' and vers == '':
	print("Error: Version must be defined in release mode")
	sys.exit()
```
This was code that was living in large_template and missed getting integrated into the original gslab_scons release. This step should involve removing this code from the template.

* Rethink how we are doing logging in a way that tracks build time and warnings across runs for each build step. See #16 for a discussion.
* See if there is a way to write unit tests for release.py functions.
[...]

PR #47, opened 2017-01-19
Pull request for #28: gslab_scons restructure

stanfordquan:
Changelog:

- [x] check_code_extension 
Done 
- [x] target_file
Done
 
- [x] informative error message
Done
 
- [x] build_tables syntax
Done.
 
- [x] restructure build_stata
Implemented.
 
- [x] start_log mode = "develop, cache, release"
I implemented this, but it's not immediately obvious to me why we want to do this, since this is code related to the GSLab template, whereas this package is supposed to contain only general-purpose builders. I feel the same about the `check_lfs()` that we do at beginning of log files.
Change is implemented in both `template` and `gslab_scons`, and it will break compatibility if users update `gslab_scons` without updating `template` starting this commit.
 
- [ ] tracks build time
Right now, tracking build time is the builder's job. It writes to stdout (and thus also the log) at the start and the end the machine's current time. 
 
- [x] unit test for release.py
done in #33
 
- [ ] cache reconfig
I read the scons-configure-cache.py and relevant man pages. The use of this file seems to be reconfiguring the cache so that the subdirectory's prefix have different lengths (e.g. 2: 00 0A 0B, 3: 000 00A 00B...). That's the only information the config file in the cache repo holds, for what it's worth. I did not implement this suggestion but we should talk more about what this is for since I may not be up to page on this.
 
- [x] env input 
As of now, my solution is to not edit the gslab_python but change the way we specify source code vs. input asset in the SConscripts themselves. I see this as the most clear solution. Runnerup solution is:
```
<SConscript>
target = '#output/plot.eps'
source = '#source/analysis.do'
assets = '#source/raw/data.txt'
env.Stata(target, source, assets)
```
where the builder definition in SConstruct is modified so that it runs
```
env.Depends(target, assets)
build_stata(target, source)
```
but this requires frontloading the Builders definitions with more scripting that isn't necessarily cleaner than my proposed solution here. Either way, there's no gslab_python code change, just convention changes in template. I could have restrict "source" to be single-element string but that would be backward incompatible which I think is unnecessary at this point for our lab and collaborators.
 
- [x] restructuring log file
Log files are now copied and appended into the overall SConstruct log in the following order:
```  
start time
builder log
builder log
Sconstruct log
```
SConstruct log also no longer overwrite with each build but rather append. I do not yet delete the SConscript.log files, which have the same behaviour as before: overwrite with each build. I figure it'll make it easier to debug specific logs, but I'm happy to simply delete the SConscript.log file after build.
I have some concerns about this implementation, in particular: accumulating records may introduce significant merge conflicts. Branches A and B from Master will have different log files after merging in. Even with timestamps, it'll make things confusing. I think that Git history tool or blame tool are appropriate for this use of searching the most recent run of a particular build, or printing out SConsign file, rather than this implementation.


arosenbe:
Request noted. Barring something new on congress_text, this is up next for me.

gslab_scons/log.py
```diff
@@ -2,27 +2,42 @@
...
     sys.stderr = sys.stdout 
+    print start_message.format("^")
+
     return None
 
 
-def log_timestamp(start_time, end_time, filename):
+def log_timestamp(start_time, end_time, filename = 'sconstruct.log'):
     '''Adds beginning and ending times to a log file.'''
     with open(filename, mode = 'r+U') as f:
```
arosenbe: Why do we use mode `'r+U'` instead of our standard `'rU'`?

arosenbe:
See this comment on the R builder in the sister issue in gslab-econ/gslab_template#27 (will link once the review comments are posted from there).

stanfordquan:
Back on this after short vacation in TRANSville.

gslab_scons/builders/build_lyx.py
```diff
@@ -1,4 +1,5 @@
 import os
+import subprocess
 import shutil
 import gslab_scons.misc as misc
```
stanfordquan: I'm agnostic about this, so I'll implement it.

stanfordquan: On second thought, I disagree. There are only three Stata only helper functions in misc; this method means we'll be importing four to five helper functions separately and I'm not convinced that will be cleaner than the current implementation.

gslab_scons/builders/build_lyx.py
```diff
@@ -21,19 +22,41 @@ def build_lyx(target, source, env):
         be the .lyx file that the function will compile as a PDF.
```
stanfordquan: added     
`env: SCons construction environment, see SCons user guide 7.2`

gslab_scons/builders/build_stata.py
```diff
@@ -31,44 +31,30 @@ def build_stata(target, source, env):
     target_file  = str(target[0])
 
     target_dir   = os.path.dirname(target_file)
```
stanfordquan: fixed

gslab_scons/builders/build_lyx.py
```diff
@@ -21,19 +22,41 @@ def build_lyx(target, source, env):
...
+    # System call
+    try:
+        command = 'lyx -e pdf2 %s > %s' % (source_file, log_file)
+        subprocess.check_output(command,
+                                stderr = subprocess.STDOUT,
+                                shell = True)
+        # Move rendered pdf to the target
+        shutil.move(newpdf, target_file)
+    except subprocess.CalledProcessError:
+        message = system_call_error("lyx", command)
```
stanfordquan: done

gslab_scons/misc.py
```diff
@@ -88,22 +133,26 @@ def make_list_if_string(source):
     return source
 
 
-def check_code_extension(source_file, software):
+def check_code_extension(source_file, extension):
     '''
     This function raises an exception if `source_file`'s extension
```
stanfordquan: done

gslab_scons/builders/build_tables.py
```diff
@@ -19,12 +19,19 @@ def build_tables(target, source, env):
...
+    misc.check_code_extension(source_file, '.lyx')
+
+    # Setup target file (filled table)
+    target_file = str(target[0])
+    misc.check_code_extension(target_file, '.lyx')
     
-    tablefill(input    = ' '.join([str(a) for a in source[1:len(source)]]), 
-              template = str(source[0]), 
-              output   = str(target[0]))
+    tablefill(input    = ' '.join([str(a) for a in source[1:]]),
```
stanfordquan: I can write `str(i) for i in source[1:]])`. Given this variable lives only on this line, I don't think it's clean to give it a longer name.

gslab_scons/log.py
```diff
@@ -2,27 +2,42 @@
 import sys
```
stanfordquan: fixed

gslab_scons/log.py
```diff
@@ -2,27 +2,42 @@
...
 
 
-def start_log(log = 'sconstruct.log'):
+def start_log(mode, vers, log = 'sconstruct.log'):
     '''Begins logging a build process'''
     check_lfs()
+    
+    if not (mode in ['develop', 'cache', 'release']):
+        print("Error: %s is not a defined mode" % mode)
+        sys.exit()
```
stanfordquan: Not really, changed.

gslab_scons/log.py
```diff
@@ -2,27 +2,42 @@
...
+    if mode == 'release' and vers == '':
+        print("Error: Version must be defined in release mode")
+        sys.exit()
+
+    start_message = "\n{0}{0}{0} New build: " + misc.current_time() + " {0}{0}{0}"
```
stanfordquan: good with me

gslab_scons/log.py
stanfordquan: Check out this wonderful chart! 
chart

gslab_scons/builders/build_lyx.py
```diff
@@ -21,19 +22,41 @@ def build_lyx(target, source, env):
...
     
-    os.system('lyx -e pdf2 %s > %s' % (source_file, log_file))
-    
-    shutil.move(newpdf, target_file)
+    # System call
+    try:
+        command = 'lyx -e pdf2 %s > %s' % (source_file, log_file)
+        subprocess.check_output(command,
+                                stderr = subprocess.STDOUT,
+                                shell = True)
```
stanfordquan: done

stanfordquan:
Remaining action items
- [ ] README update and additional documentation
- [ ] Race conditions and log refactoring
- [ ] Think about make_list_if_string()

stanfordquan:
@arosenbe what I understand from our discussion is that regarding to logging, we will:

1) keep outputting build logs with timestamp
2) reconstruct Sconstruct.log, adding to it builder logs. This is to avoid race conditions.
3) do _NOT_ append SConstruct logs for each run, instead rewriting. Same for builder logs. This is to avoid merge headaches. 

Let me know if you agree.

stanfordquan:
@arosenbe 

Regarding the concerns that were raised, here is how I chose to tackle them:

1) Tutorial on builder use: I think this is unnecessary given the current template for calling a builder:

```
source = '#source/analysis/analysis.do'
assets = '#output/data/data.txt'
env.Depends(target, assets)
env.Stata(target, source)
```
which is in response to the MG's desire to separate source code and assets in template.
In addition, please note that very little code in `gslab_python` was changed to accommodate this style (we still accept the "source code first" style). This is, IMO, so that we don't introduce unnecessary backward-incompatible changes.

2) Builder logs: The current construction is that the Sconstruct is rewritten with each run. For builders that are called during a Scons run, their Sconscript.log is appended to the Sconstruct.log using an `atexit` call, starting with most recent complete time. Since this is all `atexit` in the Sconstruct, there should be no concern regarding race conditions.

3) Make_list_if_string(): I enforce that input to Scons builder arguments are either `str` or `list`.

gslab_scons/builders/build_python.py
```diff
@@ -15,19 +16,37 @@ def build_python(target, source, env):
...
-    os.system('python %s %s > %s' % (source_file, cl_arg, log_file))
+    # System call
+    try:
+        command = 'python %s %s > %s' % (source_file, cl_arg, log_file)
+        subprocess.check_output(command,
+                                stderr = subprocess.STDOUT,
+                                shell  = True)
+    except subprocess.CalledProcessError:
+        message = misc.command_error_msg("Python", command)
+        raise BadExecutableError(message)
```
arosenbe: It doesn't look like BadExecutableError is loaded.

gslab_scons/builders/build_matlab.py
```diff
@@ -20,35 +22,48 @@ def build_matlab(target, source, env):
...
+                                shell  = True)
+    except subprocess.CalledProcessError:
+        message = misc.command_error_msg("Matlab", command)
+        raise BadExecutableError(message)    
     os.remove('source.m')
 
     end_time = misc.current_time()
     log_timestamp(start_time, end_time, log_file)
+
     return None
```
arosenbe: Carriage Return.

gslab_scons/log.py
```diff
@@ -1,28 +1,82 @@
...
+
+def collect_builder_logs(parent_dir):
+    ''' Recursively return dictionary of files ending with sconscript.log 
+        in parent_dir and nested directories.
+        Also return timestamp from those sconscript.log 
+        (snippet from SO 3964681)'''
+    builder_log_collect = {}
+
+    for root, dirs, files in os.walk(parent_dir): # take a walk in parent and child dirs
+        for file in files:
```
arosenbe: Let's not overwrite base python objects.
```python
for f in files:
```

gslab_scons/log.py
```diff
@@ -1,28 +1,82 @@
...
+            if file.endswith("sconscript.log"):
+                full_file = os.path.join(root, file)
+                with open(full_file, 'rU') as f:
```
arosenbe: We can make this f `f_full`.

gslab_scons/log.py
```diff
@@ -1,28 +1,82 @@
...
+                    s = f.readlines()[1] # skip a line
```
arosenbe: Why do we use the variable `s`?

gslab_scons/log.py
```diff
@@ -1,28 +1,82 @@
...
+def end_log(log = 'sconstruct.log'):
+    '''Complete the log of a build process.'''
+
+    end_message = "*** Build completed: {%s} ***\n" % misc.current_time()
+    with open(log, "a") as f:
+        f.write(end_message)
+
+    # scan sconstruct.log for start time (if looks unpythonic, see xkcd 1171)
+    with open('sconstruct.log', "r") as f:
+        s = f.readline()
```
arosenbe: Why do we use the variable `s` here?

gslab_scons/log.py
```diff
@@ -1,28 +1,82 @@
...
+    parent_dir = os.getcwd()
+    builder_logs = collect_builder_logs(parent_dir)
+    
+    # keep only builder logs from this run
+    this_run_dict = {key:value for key, value in builder_logs.items() if value > start_time}
+    this_run_list = sorted(this_run_dict, key=this_run_dict.get, reverse=True)
 
-def log_timestamp(start_time, end_time, filename):
+    with open('sconstruct.log', "a") as sconstruct:
+        for file in this_run_list:
```
arosenbe: Don't overwrite base python.

gslab_scons/misc.py
```diff
@@ -74,16 +123,16 @@ def stata_command_unix(flavor, cl_arg = ''):
                'linux' : '-b',
                'linux2': '-b'}
     option  = options[platform]
-    command = flavor + ' ' + option + ' %s ' + cl_arg # %s will take filename later
+    command = flavor + ' ' + option + ' %s %s'  # %s will take filename later
```
arosenbe: sprintf, please.

gslab_scons/misc.py
```diff
@@ -74,16 +123,16 @@ def stata_command_unix(flavor, cl_arg = ''):
...
 
 
-def stata_command_win(flavor, cl_arg = ''):
+def stata_command_win(flavor):
     '''
     This function returns the appropriate Stata command for a user's 
     Windows platform.
     '''
-    command  = flavor + ' /e do' + ' %s ' + cl_arg # %s will take filename later
+    command  = flavor + ' /e do' + ' %s %s'  # %s will take filename later
```
arosenbe: sprintf, please

gslab_scons/builders/build_lyx.py
```diff
@@ -19,21 +20,38 @@ def build_lyx(target, source, env):
...
     source      = misc.make_list_if_string(source)
     target      = misc.make_list_if_string(target)
     start_time  = misc.current_time()
+    
+    # Set up source file
     source_file = str(source[0])
+    misc.check_code_extension(source_file, '.lyx')
+
+    # Set up target file and log file
+    newpdf      = source_file.replace('.lyx','.pdf')
```
stanfordquan: fixed!

gslab_scons/builders/build_python.py
stanfordquan: fixed!

gslab_scons/log.py
stanfordquan: Done!

gslab_scons/log.py
stanfordquan: Good catch!

gslab_scons/log.py
stanfordquan: resolved verbally

stanfordquan: `s` short for string

gslab_scons/misc.py
stanfordquan: Resolved, @arosenbe happy to leave it the way it is.

gslab_scons/misc.py
stanfordquan: Resolved, @arosenbe happy to leave it the way it is.

arosenbe:
@stanfordquan, 

I'm getting the following error when trying to run the template. I think you're brining too much information into `s` in the logging functions.

stanfordquan:
@arosenbe , your log files do not have the timestamp as specified, probably because you had a build folder.

This release is backward incompatible due to the fact that these logs must have timestamps.

arosenbe:
Nice work @stanfordquan; good to merge.

stanfordquan:
@jmshapir @gentzkow 

For `gslab_python`, some changes are backward incompatible in the sense that it requires an updated `template` to run. 

We decided that it's good practice to direct these changes into a "dev_*" branch, where * is a new "major release" number in the semantic versioning. This seems to be what big projects do (e.g. [here](https://github.com/fifafu/BetterTouchTool/milestones) and [here](https://github.com/python/cpython)): directing issues into big milestones that are backward incompatible.  This way we don't have to make backward incompatible changes in this repo and `template` more than, say, twice a year.

jmshapir:
@stanfordquan I'm happy to defer to @gentzkow on that but it sounds reasonable to me.

Issue #45, opened 2017-01-11
Review and improve gslab_scons unit tests


M-R-Sullivan:
In #36 (specifically, in [this comment](https://github.com/gslab-econ/gslab_python/pull/36#pullrequestreview-15982005)), we noted that the unit tests for `gslab_python`'s `gslab_scons` package should be reviewed and improved. The goal of this issue is to do so. 

The tests currently take over half a minute to run. We should probably find a way to make them run faster.

M-R-Sullivan:
Update: this is under peer review at #49

M-R-Sullivan:
Still under PR at #49.

M-R-Sullivan:
Still under PR at #49.

M-R-Sullivan:
I merged this issue's changes into `master` after finishing peer review at #49.

**Summary**
- I expanded the coverage of `gslab_scons`'s tests.
- I mocked out most external dependencies of `gslab_scons`'s.
- I identified [these items](https://github.com/gslab-econ/gslab_python/pull/49#issuecomment-284099311) as potential problems with `gslab_scons` that we may want to consider addressing at some point. 
- This issue also changed how we run `gslab_python` tests. Instead of running a package's `run_all_tests.py` script, we now use `python setup.py test` at the root of the repository. This command produces a `coverage` report for our testing.

PR #49, opened 2017-01-23
Pull request for #45: Improve gslab_scons tests

M-R-Sullivan:
Summary of changes: On this branch, I...
- Replaced calls to external software with mocks to allow testing across platform types and to focus tests on the actual `gslab_scons` units under scrutiny. 
- Introduced tests demonstrating when and how functions raise errors or act in potentially surprising ways. Some of these tests exhibit functions' behaviour when provided with inputs unlike those they are intended to be used with.
- Standardised naming and import conventions across test scripts.

I noticed that the `is_exe()` function, which is defined as

```py
def is_exe(file_path):
    '''Check that a path refers to a file that exists and can be exectuted.'''
    return os.path.isfile(file_path) and os.access(file_path, os.X_OK)
```

is redundant, as  `os.access(file_path, os.X_OK)` will fail if `file_path` doesn't exist. So the `is_exe()` calls could be replaced with `os.access()` calls. Let me know if you agree with this proposal. 

I added some other tests that show functions acting in potential undesirable ways. If you'd like, you can also let me know if you think it's worth making any other improvements as a part of this pull request or elsewhere.


M-R-Sullivan:
@stanfordquan - would you mind peer reviewing my work on this branch?

stanfordquan:
Sorry just saw this, will get on it first time next week

stanfordquan:
On hold while wrapping up TRANS.

lboxell:
@stanfordquan, I'm going to re-assign this to myself to re-balance work, unless you were planning on working this today/tomorrow.

stanfordquan:
I was going to get started but I'm perfectly happy to reassign!

gslab_scons/tests/test_build_python.py
```diff
@@ -3,40 +3,124 @@
...
+        '''
+        Test that build_python() creates a log in its
+        target's directory.
+        '''
+        gs.build_python(target = './build/py.py', 
+                       source = './input/python_test_script.py', 
+                       env    = '')
+        self.check_log('./build/sconscript.log')
+
+    def check_log(self, log_path = './build/sconscript.log'):
```
lboxell: Do we want to have a set of misc functions that we load for all unit tests?

At one extreme, it seems like we do a lot of the same tests across builders (bad extension, log creation, etc) where the only differences across these items are the build function, the source, and the target. Would it make sense/be possible to have a single `test_builders.py` script that performs the same tests across all of our builders?

Using something like http://stackoverflow.com/questions/14078357/python-how-can-idynamically-create-a-function-with-a-name-based-on-a-string-re

Let me know if this isn't clear.

gslab_scons/tests/test_build_python.py
```diff
@@ -3,40 +3,124 @@
...
+        self.check_log('./sconscript.log')       
+        # Other non-string inputs should
+        with self.assertRaises(TypeError):
+            gs.build_python(None, source, env)
+        with self.assertRaises(TypeError):
+            gs.build_python(1, source, env)
+
+        #== source ===============
+        # We can have multiple sources, and we can 
+        # specify sources that don't exist...
```
lboxell: SCons should check that all sources needed to create the target exist for us.

 Is there way to test these builders within the `scons` framework? 

It seems like there are three levels of tests:
 - mock tests (system and scons independent)
 - Test os.system calls (scons independent)
 - Test within scons 

We would like the functions to work well within all three of these environments, though I don't know what that means for the optimal way to design the unit tests.

gslab_scons/tests/test_build_python.py
```diff
@@ -3,40 +3,124 @@
...
+        # console and continue. 
+        print "Expecting console output suggesting an error:\n"
+        gs.build_python(target = '', 
+                        source = ['nonexistent_data.py',
+                              './input/python_test_script.py'], 
+                         env    = None)
+        print "No longer expecting an error message.\n"
+        self.check_log('./sconscript.log')
+
+    def test_target_creation_unnecessary(self):
```
lboxell: SCons will notice that a target wasn't created successfully, but this just causes it to re-run the next time to try and build the target. Adding something that checks for targets being created properly could be nice so we don't have more 'silent' target misspecification errors. (If you agree, a new task to implement probably makes sense.)

gslab_scons/tests/test_build_python.py
```diff
@@ -3,40 +3,124 @@
...
 
 sys.path.append('../..')
-from gslab_scons import build_python
+import gslab_scons.builders.build_python as gs
 from gslab_scons._exception_classes import BadExtensionError
+from gslab_make.tests import nostderrout
 
-
-class testbuild_python(unittest.TestCase):
+class TestBuildPython(unittest.TestCase):
```
lboxell: Why don't we mock python?

gslab_scons/tests/test_build_r.py
```diff
@@ -4,35 +4,124 @@
...
-        with self.assertRaises(BadExtensionError):
-            build_r('../build/r.rds', ['bad', './input/R_test_script.R'], env)
+        with self.assertRaises(BadExtensionError), nostderrout():
+            gs.build_r(target = './build/r.rds', 
+                       source = ['bad', './script.R'], 
+                       env    = '')
+
+    @mock.patch('gslab_scons.builders.build_r.os.system')
+    def test_unintended_inputs(self, mock_system):
+        # We don't expect the env argument to affect the function's
```
lboxell: This isn't quite true with the current master of gslab_python because we pass command line arguments through the env variable.

gslab_scons/tests/test_build_python.py
M-R-Sullivan: I agree that we should create a new task to add such checks.

gslab_scons/tests/test_build_r.py
M-R-Sullivan: OK. I will pull in the current master and update the tests as is appropriate unless you object.

gslab_scons/tests/test_build_python.py
M-R-Sullivan: I think that if we'd like these tests to be proper unit tests, then the current implementation is fine (we expect a unit with which build_python() is integrated to do the check rather than build_python() itself). This sounds more like an integration test to me, though I don't know enough about testing to say that confidently. I think this relates to the point you suggest we discuss re: whether we assume everything outside the function we're testing works in a given way or whether we actually check it works with external resources.

gslab_scons/tests/test_build_python.py
M-R-Sullivan: I didn't see much to be gained in terms of speed or portability from changing the existing structure but I'm happy to change this if you think otherwise.

gslab_scons/builders/build_r.py
```diff
@@ -22,6 +22,10 @@ def build_r(target, source, env):
     source_file = str(source[0])
     target_file = str(target[0])
     target_dir  = os.path.dirname(target_file)
+
+    if target_dir == '':
```
M-R-Sullivan: Sounds good to me!

M-R-Sullivan:
@lboxell - I think that this is ready for another round of review. To summarise my changes:
- Although I chose not to make a general `test_builders.py` script, I did add functions abstracting common types of checks made in our tests. Many of these are in `_test_helpers.py`.
- I moved the mock objects' side effects into a file called `_side_effects.py`.
- I added an `--include=<PATH>` option for `python setup.py test` that allows the user to specify the files for which coverage produces reports (this is useful when working on the tests in a particular module). 
- I revised the tests so that each `gslab_scons` script except `release.py` is covered 100%.

gslab_scons/tests/test_build_lyx.py
```diff
@@ -1,47 +1,107 @@
...
+        arguments are lists
+        '''
+        mock_system.side_effect = fx.lyx_side_effect
+        target = './build/lyx.pdf'
+
+        helpers.standard_test(self, gs.build_lyx, 'lyx', 
+                              system_mock = mock_system, target = target)
+        self.assertTrue(os.path.isfile(target))
+
+    test_bad_extension = \
```
lboxell: Why not use a normal function definition here?

gslab_scons/tests/test_log.py
```diff
@@ -1,39 +1,187 @@
...
+        for log in log_names:
+            self.assertTrue(os.path.isfile(log))
+            os.remove(log)
+        sys.stdout = initial_stdout
+
+    @staticmethod
+    def check_lfs_effect(*args, **kwargs):
+        raise ex_classes.LFSError('Git LFS is not installed.')
 
     def test_log_timestamp(self):
```
lboxell: No mocking!?

gslab_scons/tests/test_build_matlab.py
```diff
@@ -1,50 +1,104 @@
...
+
+    @helpers.platform_patch('darwin', path)
+    @main_patch
+    def test_unix(self, mock_copy, mock_system):
+        '''
+        Test that build_matlab() creates a log and properly submits
+        a matlab system command on a Unix machine.
+        '''
+        # Mock copy so that it just creates the destination file
+        mock_copy.side_effect = \
```
lboxell: Is there a way to define these side effects once? (This comment applies throughout this and other scripts.)

gslab_scons/tests/test_release_tools.py
```diff
@@ -82,94 +84,106 @@ def test_up_to_date(self, mock_call):
...
+        self.assertEqual(repo_info[2], 'branch')
 
-    def test_create_size_dictionary(self):
+        # Ensure that extract_dot_git() raises an error when the directory
+        # argument is not a .git folder.
+        # i) The directory argument identifies an empty folder
+        with self.assertRaises(ReleaseError):
+            repo_info = tools.extract_dot_git('not/git')
+
+        # ii) Mock the .git/config file so that the url line doesn't
```
lboxell: I don't follow what this is supposed to be testing or why it should error.

gslab_scons/tests/test_release_function.py
```diff
@@ -40,37 +76,65 @@ def test_release_standard(self, mock_move, mock_make_archive, mock_copy,
...
-                      target_commitish  = 'test_branch', 
-                      zip_release       = True)
+        self.check_release(args, mock_session, mock_upload, mock_copy,
+                           mock_make_archive, mock_move)
+
+        # Did release() create the local_release subdirectory?
+        mock_makedirs.assert_any_call(local_release)
+        # Did it create the temporary directory for to-be-zipped release files?
+        mock_makedirs.assert_any_call('release_content')
+
```
lboxell: Do we need a patch here?

gslab_scons/tests/test_release_function.py
```diff
@@ -13,25 +17,57 @@
...
+    def test_release_standard(self, 
+                              mock_move, 
+                              mock_make_archive, 
+                              mock_copy, 
+                              mock_makedirs, 
+                              mock_getpass, 
+                              mock_session,
+                              mock_upload, 
+                              mock_rmtree,
+                              mock_isdir):
```
lboxell: Is it ok to not reference everything in the patch?

gslab_scons/tests/test_release_function.py
```diff
@@ -80,133 +144,103 @@ def test_release_standard(self, mock_move, mock_make_archive, mock_copy,
...
+        # local_release directory.
+        else:
+            base = args['local_release']
+
+        drive_files = args['DriveReleaseFiles']
+        if isinstance(drive_files, basestring):
+            drive_files = [drive_files]
+
+        for filename in drive_files:
+            mock_copy.assert_any_call(filename, '%s/%s' % (base, filename))
```
lboxell: Why not `assert_called_with`?

gslab_scons/tests/test_misc.py
```diff
@@ -1,95 +1,305 @@
...
+        # Total access except execution
+        other_files   = ['test_script.do']
+        
+        # Access the arguments of the os.access() call.
+        path = args[0]
+        mode = args[1]
+
+        if mode == os.X_OK and path in execute_files:
+            result = True
+        elif mode != os.X_OK and path in (execute_files + other_files):
```
M-R-Sullivan: `os.access()` is only ever called with `mode = os.X_OK` in `misc.py`, but I thought I'd make the mock more realistic by allowing different modes so that it would catch changes we may make (i.e. changing the the `mode` argument from `os.access()` in `misc.py`).

I am happy to get rid of this if you think it adds unnecessary complexity!

gslab_scons/tests/test_build_stata.py
```diff
@@ -3,75 +3,185 @@
...
+    @mock.patch('%s.misc.is_in_path'         % path)
+    @mock.patch('%s.subprocess.check_output' % path)
+    def test_no_executable_in_path(self, mock_check, mock_path):
+        '''
+        Test build_stata()'s behaviour when there no valid Stata
+        executables in the user's path variable
+        '''
+        # We mock the system to only find an executable called Rscript
+        # in the path.
+        mock_check.side_effect = fx.make_stata_side_effect('Rscript')
```
M-R-Sullivan: Sure!

gslab_scons/tests/test_misc.py
lboxell: I'd support reducing complexity of the tests (since there is already a lot to digest for a new user), but I don't think its a huge issue.

gslab_scons/tests/test_misc.py
```diff
@@ -120,5 +330,4 @@ def test_lyx_scan(self):
```
M-R-Sullivan: Which test did you mean to comment on?

gslab_scons/tests/test_release_tools.py
M-R-Sullivan: I will add a comment.

gslab_scons/tests/test_misc.py
lboxell: `test_lyx_scan`

gslab_scons/tests/test_release_function.py
M-R-Sullivan: All of the patches are referenced within `@patch_collection`, which I created to avoid having to include a long list of patches above each test method.

gslab_scons/tests/test_release_function.py
M-R-Sullivan: `assert_called_with()` considers only the most recent call of the mocked function, so it would fail if `len(drive_files) > 2`. See [here](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_any_call).

gslab_scons/tests/test_release_function.py
M-R-Sullivan: For `check_release()`? If so, no, as it is not a test method but a function that should receive mock objects as arguments. Let me give it a docstring...

gslab_scons/tests/test_log.py
```diff
@@ -1,39 +1,187 @@
...
+    # user has git_lfs
+
+    def setUp(self):
+        if os.path.isfile('sconstruct.log'):
+            os.remove('sconstruct.log')
+
+    # Mock a Unix platform (sys.platform = 'darwin' on  Mac machines).
+    @helpers.platform_patch('darwin', path)
+    @mock.patch('gslab_scons.log.misc.check_lfs')
+    def test_start_log_stdout_on_unix(self, mock_lfs):
```
M-R-Sullivan: As discussed, adding a `@mock.patch()` decorator without a second argument above a method passes that method a MagicMock of the object specified by the patch's first argument.  [This documentation](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) may be useful.

gslab_scons/builders/build_matlab.py
```diff
@@ -1,45 +1,51 @@
...
 
     Parameters
     ----------
     target: string or list 
         The target(s) of the SCons command.
     source: string or list
-        The source(s) of the SCons command. The first source specified
-        should be the Matlab .M script that the builder is intended to execute. 
+        The source(s) of the SCons command. The first source 
+        specified should be the Matlab .M script that the
```
M-R-Sullivan: Done.

gslab_scons/tests/_side_effects.py
```diff
@@ -0,0 +1,274 @@
...
+    # Get and parse the command passed to os.system()
+    command = args[0]
+    match   = helpers.command_match(command, 'R')
+
+    executable = match.group('executable')
+    log        = match.group('log')
+
+    if log is None:
+        # If no log path is specified, the R script's after 
+        # removing .R (if present) and adding .Rout.
```
M-R-Sullivan: Indeed!

gslab_scons/tests/_side_effects.py
```diff
@@ -0,0 +1,274 @@
...
+    match   = helpers.command_match(command, 'python')
+
+    if match.group('log'):
+        log_path = re.sub('(\s|>)', '', match.group('log'))
+        with open(log_path, 'wb') as log_file:
+            log_file.write('Test log')
+
+
+def matlab_side_effect(*args, **kwargs):
+    '''Mock os.system for Matlab commands'''
```
M-R-Sullivan: No, although I think that `build_matlab()` doesn't produce an error and produces an empty log when `matlab` isn't a recognised executable.

gslab_scons/tests/_side_effects.py
```diff
@@ -0,0 +1,274 @@
...
+            # Find the Stata script's name
+            script_name = match.group('source')
+            stata_log   = os.path.basename(script_name).replace('.do','.log')
+            
+            # Write a log
+            with open(stata_log, 'wb') as logfile:
+                logfile.write('Test Stata log.\n')
+
+        else:
+            # Raise an error if the exectuable is not the only recognised one.
```
M-R-Sullivan: Sure. I'll also fix the spelling.

gslab_scons/tests/_side_effects.py
```diff
@@ -0,0 +1,274 @@
...
+def lyx_side_effect(*args, **kwargs):
+    '''
+    This side effect mocks the behaviour of a system call.
+    The mocked machine has lyx set up as a command-line executable
+    and can export .lyx files to .pdf files only using 
+    the "-e pdf2" option.
+    '''
+    # Get and parse the command passed to os.system()
+    command = args[0]
+    match = re.match('\s*'
```
M-R-Sullivan: Good idea!

gslab_scons/tests/_side_effects.py
```diff
@@ -0,0 +1,274 @@
...
+
+    executable   = match.group('executable')
+    option       = match.group('option')
+    source       = match.group('source')
+    log_redirect = match.group('log_redirect')
+
+    option_type    = re.findall('^(-\w+)',  option)[0]
+    option_setting = re.findall('\s(\w+)$', option)[0]
+
+    is_lyx = bool(re.search('^lyx$', executable, flags = re.I))
```
M-R-Sullivan: They aren't on my computer. I'm not sure about Windows but it wouldn't be difficult for me to check this later today.

gslab_scons/tests/_side_effects.py
```diff
@@ -0,0 +1,274 @@
...
+        out_path = re.sub('lyx$', 'pdf', source, flags = re.I)
+        with open(out_path, 'wb') as out_file:
+            out_file.write('Mock .pdf output')
+
+
+def make_call_side_effect(text):
+    '''
+    Intended for mocking subprocess.call in testing
+    _release_tools.up_to_date(). Return a side effect that
+    prints text to mocked function's the stdout argument.
```
M-R-Sullivan: Indeed.

gslab_scons/tests/_side_effects.py
```diff
@@ -0,0 +1,274 @@
...
+        log_path = re.sub('>\s*', '', log_redirect)
+        with open(log_path, 'wb') as log_file:
+            log_file.write('Test log\n')
+
+    # If LyX is the executable, the options are correctly specified,
+    # and the source exists, produce a .pdf file with the same base 
+    # name as the source file.
+
+    # Make a mocked-up comprehensive list of existing files
+    existing_files = ['test_script.lyx', './input/lyx_test_file.lyx']
```
M-R-Sullivan: I revised the comment, but please let me know if this bit remains unclear.

gslab_scons/tests/_side_effects.py
```diff
@@ -0,0 +1,274 @@
...
+  def raise_http_error():
+      raise requests.exceptions.HTTPError('404 Client Error')
+
+  if release_path != real_path:
+      mock_output.raise_for_status.side_effect = raise_http_error
+
+  return mock_output  
+
+
+def dot_git_open_side_effect(repo   = 'repo',
```
M-R-Sullivan: I agree!

gslab_scons/tests/_test_helpers.py
```diff
@@ -0,0 +1,237 @@
...
+        misc_patch  = mock.patch(misc_path, platform)
+        total_patch = lambda f: misc_patch(main_patch(f))
+    except ImportError:
+        total_patch = main_patch
+
+    return total_patch
+
+def command_match(command, executable, which = None):
+    '''Parse Python, R, and Stata system calls as re.match objects'''
+    if executable in ['python', 'py']:
```
M-R-Sullivan: Good idea.

gslab_scons/tests/_test_helpers.py
```diff
@@ -0,0 +1,237 @@
...
+    source = './test_script.%s' % extension
+    target = './test_output.txt'
+
+    # i) Single command line argument
+    if env:
+        env['CL_ARG'] = 'test'
+    else:
+        env = {'CL_ARG': 'test'}
+    
+    builder(source = source, target = target, env    = env)
```
M-R-Sullivan: I fixed it.

gslab_scons/tests/_test_helpers.py
```diff
@@ -0,0 +1,237 @@
...
+    os.remove(log_path)
+
+
+def bad_extension(test_object, builder, 
+                  bad  = 'bad',
+                  good = None,
+                  env  = {}):
+    '''Ensure builders fail when their first sources have bad extensions'''
+    if good:
+        source = [bad, good]
```
M-R-Sullivan: I added `source` as an argument to the builder call and added a comment noting why I put `good` second. I was interested in checking that that `BadExtensionError`s are raised even when a source with a good extension is specified (but not in the position of the list of sources).

gslab_scons/tests/_test_helpers.py
```diff
@@ -0,0 +1,237 @@
...
+                         command)
+
+    if which:
+        return match.group(which)
+    else:
+        return match
+
+
+def check_log(test_object, log_path, timestamp = True):
+    '''Check for the existence of a timestamped log'''
```
M-R-Sullivan: OK!

gslab_scons/tests/_side_effects.py
```diff
@@ -0,0 +1,274 @@
...
+    assets_path    = kwargs['file_name']
+    shutil.copyfile(assets_path, 'assets_listing.txt')    
+
+
+def post_side_effect(*args, **kwargs):
+  '''
+  Intended for mocking requests.session.post in testing release().
+  This side effect returns a MagicMock that raises an error 
+  when its raise_for_status() method is called unless
+  a specific release_path is specified and there if a
```
M-R-Sullivan: True!

gslab_scons/tests/_side_effects.py
```diff
@@ -0,0 +1,274 @@
...
+    the only recognised system command. 
+    '''
+    def stata_side_effect(*args, **kwargs):
+        command = args[0]
+        match   = helpers.command_match(command, 'stata')
+
+        if match.group('executable') == recognised_command:
+            # Find the Stata script's name
+            script_name = match.group('source')
+            stata_log   = os.path.basename(script_name).replace('.do','.log')
```
M-R-Sullivan: Yes, thanks for that!

gslab_fill/tests/test_textfill.py
```diff
@@ -8,7 +8,7 @@
 import shutil
 
 # Ensure that Python can find and load the GSLab libraries
-os.chdir(os.path.dirname(os.path.realpath(__file__)))
+#os.chdir("gslab_fill/tests/")
```
M-R-Sullivan: We can and we should...

setup.py
```diff
@@ -1,8 +1,44 @@
...
+    include_arg = None
+
+
+class TestRepo(build_py):
+    '''Build command for running tests in repo'''
+    def run(self):
+        if include_arg:
+            coverage_command = 'coverage report -m %s' % include_arg
+        else:
+            coverage_command = 'coverage report -m --omit=setup.py,*/__init__.py'
```
M-R-Sullivan: I think so, as we test neither our tests nor `./eggs` files and the "results" for these files clutter the ultimate report. Let me know if you disagree.

README.md
```diff
@@ -38,6 +38,20 @@ which are the SSH and HTTPS protocol versions.
...
+-------
+
+We recommend that users use [coverage](https://pypi.python.org/pypi/coverage/) 
+to run this repository's unit tests. Upon installing coverage (this can be done with
+pip using the command `pip install coverage`), one may test `gslab_python`'s contents
+and then produce a code coverage report the commands:
+
+```bash
+coverage run setup.py test
+coverage report -m
```
M-R-Sullivan: Indeed.

gslab_scons/tests/test_misc.py
M-R-Sullivan: MagicMock() is used here because it is more straightforward to deploy than an SCons `node` object (which it is supposed to be mocking). This mock just gives the contents of a test LyX file when its `get_contents()` method is called. I'll add a comment.

gslab_scons/tests/test_log.py
```diff
@@ -1,39 +1,187 @@
...
+        '''
+        Test that start_log() leads stdout to be captured in 
+        a log file on Unix machines. 
+        '''    
+        initial_stdout = sys.stdout
```
M-R-Sullivan: Done.

gslab_scons/tests/test_log.py
M-R-Sullivan: Indeed... Very shocking!

gslab_scons/tests/test_log.py
```diff
@@ -1,39 +1,187 @@
...
+import gslab_scons._exception_classes as ex_classes
+from gslab_make.tests import nostderrout
+
+# Define path to the builder for use in patching
+path = 'gslab_scons.log.misc'
+
+
+class TestLog(unittest.TestCase):
+
+    # I think that we should mock check_lfs in all tests
```
M-R-Sullivan: Nope... nice catch!

gslab_scons/tests/test_build_python.py
```diff
@@ -1,50 +1,108 @@
...
+    def test_log_creation(self, mock_system):
+        '''Test build_python()'s behaviour when given standard inputs.'''
+        mock_system.side_effect = fx.python_side_effect
+        helpers.standard_test(self, gs.build_python, 'py', 
+                              system_mock = mock_system)
+
+    # Test that build_python() recognises an inappropriate file extension
+    test_bad_extension = \
+        lambda self: helpers.bad_extension(self, gs.build_python, 
+                                           good = 'test.r')
```
M-R-Sullivan: No - I changed it to `test.py`.

gslab_scons/tests/_test_helpers.py
```diff
@@ -0,0 +1,237 @@
...
+                env    = env)
+
+def standard_test(test_object, builder, 
+                  extension   = None,  
+                  system_mock = None,
+                  source = None,
+                  target = None,
+                  env    = None):
+    '''Test that builders run without errors and create logs properly.'''
+    if not source:
```
M-R-Sullivan: Good idea.

M-R-Sullivan: I made a few changes along these lines in b9f18d8, but I kept a couple `source` definitions in function bodies because they require an `extension` argument.

gslab_scons/tests/test_build_lyx.py
M-R-Sullivan: No good reason... I replaced all lambdas like this with normal function definitions.

gslab_scons/tests/test_build_r.py
```diff
@@ -1,54 +1,90 @@
...
+        mock_system.side_effect = fx.r_side_effect
+        helpers.standard_test(self, gs.build_r, 'R', 
+                              system_mock = mock_system)
+
+    @system_patch
+    def test_target_list(self, mock_system):
+        mock_system.side_effect = fx.r_side_effect
+        # We don't expect that the targets actually need
+        # to be created.
+        targets = ['./build/r.rds', 'additional_target']
```
M-R-Sullivan: Indeed. I made some changes in response to this point in a44d315.

gslab_scons/tests/test_build_matlab.py
M-R-Sullivan: I moved this side effect and a Stata side effect maker to `_side_effects.py`, which I think makes things cleaner. See a84ee20.

M-R-Sullivan:
@lboxell - I think I've responded to your points. I'll look over the tests and make a list of potential code problems found during this test review and let you look over it before merging. Let me know if you have any other comments.

gslab_scons/tests/_test_helpers.py
lboxell: I *think* you can use other arguments to define defaults, so you could move all of them to defaults. But it isn't a big deal either way.

lboxell:
@M-R-Sullivan, I looked over your replies and they seemed fine to me. I also made a small change to prevent printing coverage information regarding .eggs scripts. Once the list of suggested changes is made, I think we can close this task.

M-R-Sullivan:
Here is the list. The items are not necessarily problems but they are things we may want to reconsider at some point. @lboxell - if you're OK with this, I'll merge. 

Did we make a decision about releasing a new version whenever we update GSLab Python? I think I made backwards-incompatible changes here by renaming `misc.command_line_args()` and replacing an `elif` with an `else` in  `_release_tools`. Our testing-procedure changes are clearly backwards incompatible (I'm not sure if these are considered in semantic versioning...). If we're accumulating changes until our next release this isn't of much note. 

### Potential problems noted in test refactoring
- `build_matlab()` doesn't check that `matlab` is a recognised executable. The function doesn't raise an error if it isn't recognised; instead, the `os.system()` call just prints 
```
sh: matlab: command not found
```
- `build_python()` doesn't check that its source script exists. The function doesn't raise an error if the script doesn't exist; instead, the `os.system()` call just prints the bash message:
```
/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python: can't open file 'test.py': [Errno 2] No such file or directory
```
- I think that `subprocess` commands like `check_output()` may handle command-line errors like those above better than `os.system()` does.
- `build_tables()` doesn't check that one of its sources is a `.lyx` file (i.e. that the user specifies a template as a source).
- Not all builders have to create their targets (this isn't necessarily a problem because SCons manages target creation, but we may want builders to leave helpful messages when they don't create targets).
- Our functions that check for and set options based on platform type don't handle [platforms](https://docs.python.org/2/library/sys.html#sys.platform) other than 'win32', 'linux(2)', and 'darwin'. I don't think it's worth changing this, but I'll flag it for the record. 
- `misc.check_code_extension()` is case-sensitive in its `software`argument (so `misc.check_code_extension('test.r', 'R')` raises an error).
- (not really discovered here, but I'll include it anyway...)  `release.py` excuses files that aren't versioned from size warnings only when their paths are explicitly stated in `.gitignore` (so files captured by `.gitignore` glob expressions receive warnings).

lboxell:
@M-R-Sullivan, Looks good.

M-R-Sullivan:
Link to list [here](https://github.com/gslab-econ/gslab_python/pull/49#issuecomment-284099311) added to outline. Merging now.

Issue #34, opened 2016-11-16
Create state of repo logger


lboxell:
See https://github.com/gslab-econ/template/issues/19#issuecomment-260826983 and https://github.com/gslab-econ/template/issues/19#issuecomment-260829667. Should live in gslab_python and, by default, run after every scons run. Should be printed to sconstruct (or it's own) log file.

There should be some safety checks to ensure it doesn't take forever to run through millions of files and shouldn't do any computational steps (i.e., compute hashes).

lboxell:
Need to confirm that if you do `sconsign .sconsign.dblite` as the last step of the build process that it outputs the same thing as doing it after the build process has finished (up to being self-documenting).

lboxell:
Note that .sconsign.dblite is edited after the build finishes, so `sconsign .sconsign.dblite` isn't that informative.

lboxell:
@gentzkow @jmshapir @arosenbe @stanfordquan @yuchuan2016 @M-R-Sullivan:

The current output for the state_of_repo logger looks like [this](https://github.com/gslab-econ/template/blob/e02ee63f7bd7688cb7964570d7dd46bcd6a0f07e/state_of_repo.log). 

I currently print two things: (1) GIT STATUS (last git commit and files that have been modified since then) and (2) The modification time and size of all files at or below the root directory where scons was run from (with a limit of 10 files per subdir). 

I'm not sure there is a way to use information in the .sconsign.dblite because this seems to get edited after all the build steps have finished (unless we make the state_of_repo.log creation separate from the main `scons` run).  

The state_of_repo.log gets completely re-written after each run. 

One thing that currently can't be determined from this is the scons tree. It looks like it will take a bit of work to print the scons tree when scons is actually running (instead of doing it manually after the run), but I can try and do this if people think it will be useful.

Let me know if you guys have any thoughts on changing/improving this.

jmshapir:
@lboxell this looks to me like the right balance of effort and reward for now. If printing the tree is very easy it would be a nice addition but it's also fine with me to leave it on the to-do list and move forward with this.

gentzkow:
I think this is a good place to stop on this for now.
-- 

Matthew Gentzkow
Professor of Economics
Stanford University

lboxell:
Closing task.

Summary:
 - Created new [release](https://github.com/gslab-econ/gslab_python/releases/tag/v3.0.3) that implmements a state of repo logger.
[...]

PR #51, opened 2017-01-25
Pull request for #34: Adds State of Repo Logger

lboxell:
@yuchuan2016, can you PR this and https://github.com/gslab-econ/template/pull/32 together?

One thing I was unsure about is whether the state_of_repo logger should go in misc.py or the builders. Technically, we use it as a builder, but it really is more of a 'misc' type of function that we use and don't expect the end user to modify.


yuchuan2016:
@lboxell , nice work! I think it's fine to store it in `misc`. Do you think it would be helpful to add `maxit` as an option that allows users to specify?

lboxell:
@yuchuan2016,

I added the maxit option and updated the template accordingly. I also merged master into this branch. Let me know if there is anything else.

yuchuan2016:
Hi @lboxell , I run the template on my PC, and files like `.\.gitattributes` are also recorded. Would you add
`re.search('\.\\.', os.path.join(root, name))` besides `re.search('\./\.', os.path.join(root, name))`?

lboxell:
@yuchuan2016,

I made the edit. Let me know if it works now.

yuchuan2016:
@lboxell , it works! Feel free to merge.

Issue #53, opened 2017-01-28
Incorporate matlab builder


jmshapir:
As far as I know we don't yet have a matlab builder for scons. I wrote a rudimentary one for my course. 
It is attached here: [build_matlab.txt](https://github.com/gslab-econ/gslab_python/files/737134/build_matlab.txt)

The purpose of this issue is to improve it and fold it into our library.

I've assigned to everyone but feel free to shuffle around as you like.

lboxell:
Closing task.

Summary:
 - Created matlab builder and tested on mac. Jesse will test on matlab at some later point and open up a task if there are issues.
 - New release: https://github.com/gslab-econ/gslab_python/releases/tag/v3.0.4

PR #54, opened 2017-01-28
Pull request for #53: Adds Matlab builder

lboxell:
@stanfordquan, I'm not really sure how to make matlab not print results to the screen, but I somewhat think its fine. For example, if a matlab script errors, I think it just hangs and doesn't close unless you close it manually. As such, being able to see the errors on the screen may be nice. Let me know if this doesn't agree with your matlab experience.

Once we finalize this on mac, we can have Jesse test on windows (because I don't think we don't have matlab access on any of our windows machines).


stanfordquan:
On it!

gslab_scons/builders/build_matlab.py
```diff
@@ -0,0 +1,53 @@
...
+
+    shutil.copy(source_file, 'source.m')
+    command  = 'matlab %s -r source -logfile %s'
+    os.system(command % (options, log_file))
+    os.remove('source.m')
+
+    end_time = misc.current_time()
+    log_timestamp(start_time, end_time, log_file)
+    return None
+
```
lboxell: Another one???

gslab_scons/builders/build_matlab.py
```diff
@@ -0,0 +1,53 @@
...
+	'''
+    start_time =  misc.current_time()
+
+    if misc.is_unix():
+        options = '-nosplash -nodesktop'
+    elif misc.is_64_windows():
+        options = '-nosplash -minimize -wait'
+    else:
+        raise Exception("Unknown OS")
```
lboxell: Thanks!

gslab_scons/builders/build_matlab.py
```diff
@@ -0,0 +1,53 @@
...
+    log_file     = target_dir + '/sconscript.log'
+    cl_arg       = misc.command_line_arg(env)
+
+    os.environ['CL_ARG'] = cl_arg
+
+    shutil.copy(source_file, 'source.m')
+    command  = 'matlab %s -r source -logfile %s'
+    os.system(command % (options, log_file))
+    os.remove('source.m')
+
```
lboxell: Matlab doesn't seem to like running things not in its path or in the same directory (https://www.mathworks.com/help/matlab/matlab_env/startup-options.html). There might be some way around this, but I'm not sure its worth the time investment.

gslab_scons/__init__.py
```diff
@@ -13,4 +13,5 @@
 import misc
 from .log import start_log, log_timestamp
 from . import builders
-from .builders import build_r, build_lyx, build_stata, build_tables, build_python
+from .builders import build_r, build_lyx, build_stata, build_tables, build_python, build_matlab
+
```
lboxell: Done.

gslab_scons/builders/__init__.py
```diff
@@ -10,3 +10,4 @@
 from .build_tables import build_tables
 from .build_python import build_python
 from .build_stata  import build_stata
+from .build_matlab import build_matlab
```
lboxell: Done.

lboxell:
@stanfordquan,

I believe I've addressed your comments. I don't think we need to add every new builder to the template. I think we want to keep the requirements of running the template to a minimum and matlab seems to be like a particularly prohibitive requirement. The unit test environment should allow you to do some testing of the builder.

Let me know if there is anything else.

gslab_scons/builders/build_matlab.py
stanfordquan: Oh sorry! Saw the red github sign of "end of file" and mistook that as "no new line"!

gslab_scons/builders/build_matlab.py
stanfordquan: All good

lboxell:
@jmshapir,

We don't have a windows machine with matlab installed to test out the matlab builder on windows. Here's some instructions you can do to test things out yourself on windows:

``` 
git clone -b issue53_matlab git@github.com:gslab-econ/gslab_python
cd gslab_python
python setup.py install clean 
cd gslab_scons/tests
pip install mock
python run_all_tests.py
```

stanfordquan:
Good to merge as long as @jmshapir  confirms things work as intended, sorry for the delay. 

One final thing: the builder prints out Matlab's console to the command line, while I don't think this is expected behaviour for the other builders. It's fine with me, but if is unintended behaviour then we probably should use `>`  to direct it to log instead. 

I like the behaviour because then you can selectively silence lines in `.m` files like we do using semi-colons.

lboxell:
@stanfordquan: Fixed the log printing.

jmshapir:
@stanfordquan I won't have time to test this for a bit so rather than hold up the pull I suggest you proceed. When I get a chance I will install this and give it a go and will open a new issue if I have trouble.

Note that you might be able to get a trial edition of Matlab on a windows machine for testing purposes.

PR #55, opened 2017-02-03
Pull request for part of #45: Refactors testing run

lboxell:
@M-R-Sullivan,

I played around with re-structuring how we run our unit tests. I think this matches more closely with what people do in the "real world" (see [here](http://doc.pytest.org/en/latest/goodpractices.html)). 

Also, I think we should start using the [coverage.py](http://coverage.readthedocs.io/en/latest/source.html) tool for part of our testing (installed via `pip install coverage`). 

We can then follow [this](https://github.com/audreyr/how-to/blob/master/python/use_coverage_with_unittest.rst), by:

```
coverage run setup.py test
coverage report -m
```

which will run all our unit tests and report the source code coverage. You can also run all of the unit tests by simply doing `python setup.py test` now.

(I also deleted the gslab_make test suite except `nostderrout.py`, because it wasn't running without errors in the new framework and I didn't think it worthwhile to maintain at this point.)

Let me know your thoughts on this setup and if we should integrate as part of #45 or not.


lboxell:
Or even better, check branches via:
```
coverage run --branch setup.py test
coverage report -m
```

M-R-Sullivan:
@lboxell - my initial reaction is that this approach seems very nice and should be integrated into #45. I'll take a longer look at your changes and then pull this branch into `issue45-scons` after I'm doing merging `master` into that branch.

M-R-Sullivan:
@lboxell: 
- Did you have any logging mechanism in mind for use with pytest? 
- This generally looks good to me. One test is failing for me but I can deal with that after merging this branch into `issue45-sconstests`.
- I renamed the test gencat subclass used in the gencat tests to eliminate pytest's warning about not being able to collect them as testing classes (I think it considers all classes beginning with "test" to be test classes...).

lboxell:
@M-R-Sullivan,

 - I wasn't sure about this. I thought about manually adding a `>`, but that won't be easy to enforce/remember. I'll look into some ways to log these within the setup.py framework.
 - Sounds good.
 - Sounds good.

M-R-Sullivan:
@lboxell - OK! Merging now.

PR #56, opened 2017-02-03
Pull request for #45: Small edits to test logging

lboxell:
@M-R-Sullivan,

Here is my proposal for how we do the logging. Now a user can just type `python setup.py test` and it does the testing/coverage reporting for them in `test.log`.


M-R-Sullivan:
@lboxell - I like this approach, though I thought it would be best to use a set-up tools command class for managing testing. I also think that we could consider changing the `coverage report` call to include other options, though I think this could be considered within the main #45 branch. Let me know what you think of my changes. If they look fine to you we can merge.

lboxell:
@M-R-Sullivan, Looks good!

Issue #59, opened 2017-03-28
Fix gslab_scons.release.py's size warnings 


M-R-Sullivan:
Our release.py script doesn't ignore non-versioned files when issuing size warnings if they are not explicitly listed in the .gitignore. The goal of this issue is to edit the script so that it ignores files in directories listed in the .gitignore and whose paths are captured by glob patterns in the .gitignore. 

Also, from the outline:
> It also would be nice if it recognized that the files were already versioned, and not ask about those, but only ask about new versioned files (need to think about this part some more though).

M-R-Sullivan:
Update: under PR at #60.

M-R-Sullivan:
See this issue's pull request at #60.

M-R-Sullivan:
This passed PR at #60.

**Summary**:
- Fixed an issue causing size warnings to be issued for unversioned files.
- Removed size warnings from release.py and added them to the template in gslab-econ/template#42.

Issue #58, opened 2017-03-25
Prepare for 4.0 release 


stanfordquan:
This task will prepare a [changelog](http://keepachangelog.com/en/0.3.0/) for `gslab_python` and `template` 4.0 release. In particular:

1) Document new & removed features, bugs fixed, and changes in existing functionality. Document where they happened. 

2) Document required changes to existing projects' templates due to backward incompatibility.

3) Ping PIs for review.

lboxell:
@stanfordquan,

Some thoughts on this:
 (1) I think our pull-request + squash & merge protocol should take care of this for us. As they will document what was added and when for the major changes (and I'm not sure we want to be documenting the smaller changes). 
 (2) I think the same thing in (1) would apply here too. So, we have some dev-4 template branch, and then that merge into the master branch of template should be the record of what should change for projects using previous templates too. The only exception is features that are backwards incompatible and are not used in template; not sure what the best thing to do in this case is, but perhaps its something we decide as a part of (3) instead of before (3).
 (3) Ping after #59 and we confirm template and gslab_python branches are in-sync?

stanfordquan:
On hold until #59 is done

stanfordquan:
On hold to wrap up Politext

lboxell:
@stanfordquan @arosenbe @M-R-Sullivan @Shun-Yang,

If you don't have other tasks to do, it would probably be good to do no more than an hour of testing of the dev_four branches of gslab_python and template (they should be tested together). That way, we confirm there are no major bugs before MG/JMS review. If you see things that you think could be improved, but aren't bugs, it would probably be worthwhile to note them here too.

gentzkow:
Thanks guys.

A standard we want to hit for the template is that it runs "off the shelf" robustly and with as little setup as possible required. I.e., if we send it to a new user whose machine is not configured to our setup, the setup is simple and clear, and the probability that they get an error the first time they run it is maximized.

I therefore think it would be valuable for you guys to:

1. Set up clean environments w/ no prior setup so you can test the full setup and run process. We should do this for Mac, Linux, and Windows.

2. Devote some effort to thinking about whether there are setup steps we can remove / make optional, ways to improve the instructions in the readme, and anything else that can make the setup more robust.

3.  Spend a little time doing tire kicking to try to find the weak spots that are most likely to produce errors.

Presumably we should make this a separate task to be completed before this task is wrapped up.

stanfordquan:
[...]

PR #60, opened 2017-04-07
Pull request for #59: Fix size warnings

M-R-Sullivan:
For #59.


M-R-Sullivan:
@lboxell - thanks for PRing this. I merged master into dev_four with [this commit](https://github.com/gslab-econ/gslab_python/commit/b1d0ffffef66110326405d7b3cd17073e2a30391); it may be a good idea for you to review the ranges I made there, too.

lboxell:
Note: I changed base back-and-forth to refresh the file change comparisons.

gslab_scons/_release_tools.py
```diff
@@ -238,6 +238,70 @@ def up_to_date(mode = 'scons', directory = '.'):
...
+    except ValueError:
+        # If the git status message lists no ignored files, return empty list
+        return []
+
+    # Use normpath() so paths can be compared as strings
+    dir_path = os.path.normpath(dir_path)
+    ignore_dirs  = []
+    ignore_files = []
+    for line in message[ignored_line:len(message)]:
+        # Determine if line could be in path's directory
```
lboxell: I don't quite follow this comment. The phrase "path's directory" is unclear to me.

gslab_scons/_release_tools.py
```diff
@@ -238,6 +238,70 @@ def up_to_date(mode = 'scons', directory = '.'):
     return bool(result)
 
 
+def list_ignored_files(dir_path = './release'):
+    '''List files ignored by git 
+
+    This function returns a list of the files in the directory
+    given by the dir_path argument that are ignored by git.
```
lboxell: I'm not sure it's really necessary for this function to restrict attention to the files in a single directory. I think it would suffice to just get a normalized list of all files that are ignored and then if a file is in this ignore list, we don't issue a size warning.

gslab_scons/_release_tools.py
```diff
@@ -238,6 +238,70 @@ def up_to_date(mode = 'scons', directory = '.'):
...
+    The filenames are the dictionary's keys; their sizes in 
+    bytes are its values. 
+    '''
+    size_dictionary = dict()
+
+    if not os.path.isdir(path):
+        raise ReleaseError("The path argument does not specify an "
+                           "existing directory.")
+
+    for root, directories, files in os.walk(path):
```
lboxell: `directories` is never used?

gslab_scons/release.py
```diff
@@ -97,10 +45,61 @@
...
+    '''Issue warnings if the files versioned in release are too large'''
+
+    # Compile a list of files in /release/ that are not versioned.
+    ignored = rtools.list_ignored_files()
+    release = rtools.create_size_dictionary('./release')
+    release = {k: release[k] for k in release.keys() if k not in ignored}
+    versioned = dict()
+
+    for file_name in release.keys():
+        versioned[file_name] = release[file_name]
```
lboxell: The creation of `versioned` seems unnecessary to me. It seems like it contains the exact same information as `release`.

gslab_scons/release.py
```diff
@@ -97,10 +45,61 @@
...
+        raise ReleaseError('SCons targets not up to date.')
+    elif not rtools.up_to_date(mode = 'git'):
+        print "Warning: `scons` has run since your latest git commit.\n"
+        response = raw_input("Would you like to continue anyway? (y|n)\n")
+        if response in ['N', 'n']: 
+            sys.exit()
+
+
+def issue_size_warnings(file_MB_limit, total_MB_limit, bytes_in_MB):
+    '''Issue warnings if the files versioned in release are too large'''
```
lboxell: I'm a little confused about the role of this function. At this point, everything in release that is going to be versioned is already versioned? Really, we want these warnings to be printed at the `git commit` stage, which is before the `release` stage. Nothing new can actually be versioned in the release stage can it?

gslab_scons/release.py
```diff
@@ -1,70 +1,19 @@
 import re
 import os
 import sys
-from _release_tools     import (up_to_date, 
-                                create_size_dictionary,
-                                extract_dot_git, 
-                                release)
+import _release_tools as rtools
```
lboxell: rtools sounds like it relates to R.

gslab_scons/_release_tools.py
M-R-Sullivan: Removed directory

gslab_scons/_release_tools.py
M-R-Sullivan: I removed this comment and the code beneath it in light of my removal of the directory argument.

gslab_scons/_release_tools.py
M-R-Sullivan: Replaced with _

gslab_scons/release.py
M-R-Sullivan: Dropped `as rtools`

gslab_scons/release.py
M-R-Sullivan: Everything that is going to be versioned has been versioned, yes. I think the idea is to remind the user that they have probably done something wrong since the last release, which may be better than never issuing size warnings. I agree it would be better, in principle, to perform this check at the commit-stage. When I originally introduced the size warnings, I was mostly following [this comment](https://github.com/gslab-econ/template/issues/19#issuecomment-262024738) and some other notes in its issue.

gslab_scons/release.py
M-R-Sullivan: Good catch. I think the distinction was once relevant but that I forgot to update the code when it became irrelevant.

gslab_scons/release.py
lboxell: What about porting this functionality to be run as the last step in scons? So we run scons and populate release, and at the last moment, we spit out a warning regarding file size. I think that would make it closer to the decision point of whether to commit large files or not.

M-R-Sullivan:
@lboxell - I think that this is ready for another check. I moved the size warning code to its own script, so the release script will no longer issue size warnings. I propose adding a note to the template outline to incorporate the size-warning functionality into the template once we release version 4.0.0.

lboxell:
@M-R-Sullivan,

There is a dev-four template [branch](https://github.com/gslab-econ/template/tree/dev_four) that we can update to include this functionality. The goal is to have the template and gslab_python dev_four branches in-sync before pinging the PIs for review in #58. 

(I'll review your changes shortly)

M-R-Sullivan:
@lboxell - OK. In that case, I'll make a size-warning implementation branch off that branch.

gslab_scons/tests/test_size_warning.py
```diff
@@ -0,0 +1,312 @@
...
+        mock_isdir.side_effect   = isdir_dict_side_effect        
+        mock_walk.side_effect    = walk_dict_side_effect
+        mock_getsize.side_effect = getsize_dict_side_effect 
+
+        # Test when one directory is provided
+        sizes = sw.create_size_dictionary(['test_files'])
+
+        self.assertEqual(len(sizes), 3)
+
+        # Check that test.txt and test.jpg are in the dictionary
```
lboxell: Update comment.

gslab_scons/tests/test_size_warning.py
```diff
@@ -0,0 +1,312 @@
...
+        root_path = [k for k in sizes.keys() if re.search('root_file.txt$', k)]
+        pdf_path  = [k for k in sizes.keys() if re.search('test.pdf$', k)]
+        self.assertTrue(bool(root_path))
```
lboxell: I might prefer a length check versus a boolean, as it seems odd to do boolean on a list. Applies throughout.

gslab_scons/tests/test_size_warning.py
```diff
@@ -0,0 +1,312 @@
...
+    path = args[0]
+
+    # Mock two directories at the root, one of which has a subdirectory
+    roots       = ['test_files',      'test_files/size_test',   'release']
+    directories = [['size_test'],     [],                       []]
+    files       = [['root_file.txt'], ['test.txt', 'test.pdf'], ['output.txt']]
+
+    for i in range(len(roots)):
+        if re.search('^%s' % path, roots[i]):
+            yield (roots[i], directories[i], files[i])
```
lboxell: Is this so `path = test_files` also returns the contents of `test_files/size_test`? A comment might be good.

gslab_scons/tests/test_size_warning.py
```diff
@@ -0,0 +1,312 @@
...
+    return isfile
+
+
+#== Side effects for testing create_size_dictionary() ===
+# These functions mock a directory containing files of various sizes
+# and a system that recognises no directory other than that one.
+def isdir_dict_side_effect(*args, **kwargs):
+    '''
+    Mock os.path.isdir() so that it only recognises ./test_files/
+    as an existing directory.
```
lboxell: Update comment.

gslab_scons/tests/test_size_warning.py
```diff
@@ -0,0 +1,312 @@
...
+    for k in struct.keys():
+        file_list += [os.path.join(k, f) for f in struct[k]]
+
+    isfile = (os.path.relpath(path) in map(os.path.relpath, file_list))
+    return isfile
+
+
+#== Side effects for testing create_size_dictionary() ===
+# These functions mock a directory containing files of various sizes
+# and a system that recognises no directory other than that one.
```
lboxell: Update comment.

gslab_scons/tests/test_size_warning.py
```diff
@@ -0,0 +1,312 @@
...
+    '''
+    path = args[0]
+    if not isinstance(path, str):
+        raise TypeError('coercing to Unicode: need string or buffer, '
+                        '%s found' % type(path))
+    return path in ['test_files', 'release']
+
+
+def walk_dict_side_effect(*args, **kwargs):
+    '''Mock os.walk() for a mock directory called ./test_files/.'''
```
lboxell: Update comment.

gslab_scons/tests/test_size_warning.py
```diff
@@ -0,0 +1,312 @@
...
+
+    for i in range(len(roots)):
+        if re.search('^%s' % path, roots[i]):
+            yield (roots[i], directories[i], files[i])
+
+
+def getsize_dict_side_effect(*args, **kwargs):
+    '''
+    Mock os.path.getsize() to return mock sizes of files in our
+    mock ./test_files/ directory.
```
lboxell: Update comment.

gslab_scons/tests/test_size_warning.py
```diff
@@ -0,0 +1,312 @@
...
+        with self.assertRaises(ReleaseError), nostderrout():
+            sizes = sw.create_size_dictionary(['nonexistent_directory'])
+        # The path argument must be a string
+        with self.assertRaises(TypeError), nostderrout():
+            sizes = sw.create_size_dictionary([10])
+
+
+#== Side effects for testing list_ignored_files() ===
+# Define the mock file structure for testing list_ignored_files()
+struct = {'.': ['untracked.txt', 'make.log', 'make.py'],
```
lboxell: Is the '.' ever used?

gslab_scons/tests/test_size_warning.py
M-R-Sullivan: The check tries to ensure that walking `release` won't provide info about `test_files` and vice versa. Comment added.

gslab_scons/tests/test_size_warning.py
M-R-Sullivan: I will replace these sorts of calls with length checks where I find them.

gslab_scons/tests/test_size_warning.py
```diff
@@ -0,0 +1,312 @@
...
+
+def walk_ignored_side_effect(*args, **kwargs):
+    '''Mock os.walk() for testing list_ignored_files()'''
+    path = args[0]
+    path = os.path.relpath(path)
+
+    if path not in struct.keys():
+        raise StopIteration
+
+    if path == 'release':
```
M-R-Sullivan: You're right. I never caught this because it doesn't matter for the mock in question (we never walk `.`) but I'll correct it now.

gslab_scons/tests/test_size_warning.py
M-R-Sullivan: And, yes, I don't just check equality so `path = test_files` also yield `test_files/size_test`.

gslab_scons/size_warning.py
```diff
@@ -0,0 +1,111 @@
...
+        return []
+
+    # Loop through the listing of ignored paths, classifying them
+    # as files or directories
+    ignore_dirs  = []
+    ignore_files = []
+    for line in message[ignored_line:len(message)]:
+        as_path = os.path.normpath(line)
+
+        # Only check for ignored files in the specified "look_in" directories
```
M-R-Sullivan: I just wasn't sure that we'd want to wait  to list obviously unversioned files (below is for `congress_text`).
```py
>>> import gslab_scons
>>> import time
>>> t = time.time(); build_files = gslab_scons.size_warning.list_ignored_files('build'); elapsed = time.time() - t
>>> print elapsed
42.23977017402649
```
Forty-two seconds isn't too long, so I'm happy to reconsider if you'd like.

M-R-Sullivan:
@lboxell - I made some additional changes since your last review. Namely, I added a SCons builder that runs the state-of-repo logger and `issue_size_warnings()`. I also updated the tests, and took the user input and `sys.exit()` bits out of the size warnings. 

Two notes:
- I can remove the `look_in` options if you'd like.
- I could also add an option, as a part of this issue or in another issue, that allows the user to turn off warnings for certain files (i.e. so warnings aren't issued after every SCons run when the user doesn't care that a file is too large).

gslab_scons/tests/test_size_warning.py
```diff
@@ -228,12 +278,28 @@ def walk_ignored_side_effect(*args, **kwargs):
     if path not in struct.keys():
         raise StopIteration
 
-    if path == 'release':
+    if path == '.':
```
lboxell: Are these 'subdirs' used anymore?

gslab_scons/tests/test_size_warning.py
```diff
@@ -260,38 +326,40 @@ def isfile_ignored_side_effect(*args, **kwargs):
...
     path = args[0]
     if not isinstance(path, str):
         raise TypeError('coercing to Unicode: need string or buffer, '
                         '%s found' % type(path))
-    return path in ['test_files', 'release']
+    acceptable = ['test_files', 'test_files/size_test', 'release', '.']
+    return os.path.normpath(path) in acceptable
 
 
 def walk_dict_side_effect(*args, **kwargs):
```
lboxell: Seems like we have two pretty similar walk side effects. If its easy to combine, we should. If not, not a big deal.

gslab_scons/tests/test_size_warning.py
```diff
@@ -88,24 +121,36 @@ def test_list_ignored_files(self, mock_check, mock_walk,
...
+        # The root
+        look_in = ['.']
+        ignored = sw.list_ignored_files(look_in)
+        expect_ignored = ['root_ignored.txt',
+                          'raw/large_file.txt',
+                          'release/.DS_Store',
+                          'release/subdir/ignored.txt']
+
+        self.assertEqual(len(ignored), len(expect_ignored))
+        self.assertEqual(ignored[0], expect_ignored[0])
```
lboxell: I like the `for i in range(len(ignored)):` check that we do in lines 132-133 a bit better.

lboxell:
Lets add your point 2, to the outline and see if we feel like it would be useful after we have some more experience.

gslab_scons/size_warning.py
```diff
@@ -89,6 +78,16 @@ def list_ignored_files(look_in):
     return ignore_files
 
 
+def _is_subpath(inner, outer):
```
M-R-Sullivan: I couldn't find any standard way to do this in Python in my brief search.

gslab_scons/tests/test_size_warning.py
M-R-Sullivan: OK!

gslab_scons/size_warning.py
```diff
@@ -20,24 +20,18 @@ def issue_size_warnings(look_in = ['source', 'raw', 'release'],
         limit = file_MB_limit * bytes_in_MB
 
         if size > limit and file_name:
-            print "Warning: the versioned file %s is larger than %d MB.\n" \
+            print "Warning: the versioned file %s is larger than %f MB." \
```
M-R-Sullivan: Agreed!

gslab_scons/tests/test_size_warning.py
M-R-Sullivan: I combined them. I think the mocking in this script could be further polished, although I'm unsure whether it's worth spending time to make improvements at this point.

gslab_scons/tests/test_size_warning.py
M-R-Sullivan: Nay. I've deleted them.

gslab_scons/misc.py
```diff
@@ -4,13 +4,28 @@
...
-def state_of_repo(target, source, env):
+def scons_debrief(target, source, env):
+    '''Execute functions after SCons has built all targets'''
+    # Log the state of the repo
     env['CL_ARG'] = env['MAXIT']
     maxit = int(command_line_args(env))
+    state_of_repo(maxit)
+
+    # Issue size warnings
+    look_in = env['look_in']
```
M-R-Sullivan: I vote for keeping it explicit for now.

M-R-Sullivan:
@lboxell - I think I've responded to each of your comments and added my second point to the [outline](https://github.com/gslab-econ/gslab_python/wiki/Outline). I also added a little function for making the "Warning" text red and bold (I can get rid of this if you'd like). Let me know what you think.

lboxell:
@M-R-Sullivan, Looks good!

M-R-Sullivan:
Great. Merging now.

PR #65, opened 2017-04-24
Pull request for #58: Prepare for 4.0

stanfordquan:
What I did in this PR:

- add master to pip install git+ssh://git@github.com/gslab-econ/gslab_python.git and other instructions 
- The R builder is failing in Windows. Command line argument issue. Check if this hold true for others.
- Add setuptools as a requirement, and suggests pip as an installation tool.
- kinder time-log failure

### DIFF
--- a/gslab_scons/log.py
+++ b/gslab_scons/log.py
@@ -30,7 +30,7 @@ def start_log(mode, vers, log = 'sconstruct.log'):
 def end_log(log = 'sconstruct.log'):
     '''Complete the log of a build process.'''
 
-    end_message = "*** Build completed: {%s} ***\n" % misc.current_time()
+    end_message = "*** Build completed: {%s} ***\n \n \n" % misc.current_time()
     with open(log, "a") as f:
         f.write(end_message)
 
@@ -44,13 +44,17 @@ def end_log(log = 'sconstruct.log'):
     parent_dir = os.getcwd()
     builder_logs = collect_builder_logs(parent_dir)
     
-    # keep only builder logs from this run
-    this_run_dict = {key:value for key, value in builder_logs.items() if value > start_time}
+    # keep only builder logs from this run OR is broken (value == beginning_of_time)
+    beginning_of_time    = datetime.min # to catch broken logs (see collect_builder_logs)
+    this_run_dict = {key:value for key, value in builder_logs.items() if (value > start_time) or value == beginning_of_time}
     this_run_list = sorted(this_run_dict, key=this_run_dict.get, reverse=True)
 
     with open('sconstruct.log', "a") as sconstruct:
         for f in this_run_list:
             with open(f, 'rU') as sconscript:
+                if this_run_dict[f] == beginning_of_time:
+                    sconstruct.write("*** Warning!!! Doesn't look the sconscript below finished.\n")
+                sconstruct.write(f + "\n")
                 sconstruct.write(sconscript.read())
 
     return None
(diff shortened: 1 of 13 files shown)


stanfordquan:
Let's hold on for this PR? I'll get started on the dropping Google Drive default support.

lboxell:
I'm going to close this and we can re-open when ready.

stanfordquan:
@arosenbe , let's do this PR before pulling `master` into `dev_four` and then finish #68 and then we can pull `dev_four` into `master` and make a release!

gslab_scons/misc.py
```diff
@@ -230,10 +230,10 @@ def check_code_extension(source_file, extension):
 
 def command_error_msg(executable, call):
     ''' This function prints an informative message given a CalledProcessError.'''
-    return '''Could not call %s.
-              Please check that the executable, source, and target files
-              are correctly specified. 
-              Command tried: %s''' % (executable, call) 
+    return '''%s did not run successfully.
+    Please check execcutable/source/target are correctly specified. 
+    Check SConstruct.log for errors.
```
arosenbe: Would you object to putting the indents back?

gslab_scons/log.py
```diff
@@ -78,7 +82,11 @@ def collect_builder_logs(parent_dir):
...
                     s = g.readlines()[1] # skip a line
                     s = s[s.find('{')+1: s.find('}')] # find {} time identifier 
-                    builder_log_end_time = datetime.strptime(s, "%Y-%m-%d %H:%M:%S")
-                builder_log_collect[f_full] = builder_log_end_time
+                    try:
+                        builder_log_end_time = datetime.strptime(s, "%Y-%m-%d %H:%M:%S")
+                    except ValueError: # if the code breaks, there's no time identifier
+                        beginning_of_time    = datetime.min
+                        builder_log_end_time = beginning_of_time
+                builder_log_collect[f_full]  = builder_log_end_time
```
arosenbe: Maybe I'm not seeing it, but is this code too redundant? Can't we equivalently write

```python 
 builder_log_collect[f_full] = datetime.min
```

Also, what's the rationale for using this specific failure time?

gslab_scons/misc.py
stanfordquan: done, had it on a local commit

gslab_scons/log.py
stanfordquan: It's a way to catch broken logs. So there are two behaviors associated with your code failing. 

1) Your code fails within the process. The process still finish, but your code is broken. In that case the `atexit` logging functionality works: your log has two timestamps, and `scons` think the action completed properly while in reality it hasn't. [Here is an example.](https://gist.github.com/stanfordquan/5cf8d25f8e23f18a513f8bd040f6fe3e) You see that Scons has no idea Stata failed and only stopped the build when it couldn't find the output table in the next step.

2) Your process fails (your executable isn't there, or the process nopes out when it gets an error). This is the case for calling Python / calling. `gslab_scons` recognize this as an `ExecCallError`. Since the process returns an error, scons terminates and the log is incomplete.  In that case do this "hack" where we pretend that the log's end time is a fixed date that we know for a fact cannot possibly be. That is so that in the log-printing function we only need to print two kinds of logs: a) the ones that have an end time after this Scons run starts (i.e. of this current run) b) and the ones that broke.

stanfordquan:
I find the Sublime Text syntax highlighting for `gslab_scons` logs a bit too avant-garde for this PR and thus plan to hide it in the next release of `template`.

gslab_scons/log.py
stanfordquan: @arosenbe ok-ed this offline

Issue #43, opened 2017-01-08
Dependency on package "requests"


jmshapir:
@lboxell @arosenbe @M-R-Sullivan @stanfordquan it looks like gslab-python depends on a package called "requests" that is not automatically installed by ```setup.py```.

Is this intended behavior?

arosenbe:
A later commit overwrote the change I made in this issue: reopening to reimplement the change. See [here](https://github.com/gslab-econ/template/pull/26#issuecomment-274103638) for context.

arosenbe:
Now under PR at #48.

arosenbe:
Summary comment
  *  Add `requests` module as a dependency in `setup.py`.
  *  Ensure no other undocumented dependencies exist. 

  *  Readd `requests` module as a dependency in `setup.py` after being overwritten by later commit.

PR #66, opened 2017-04-28
Pull request for #43 in Template: Configuration tests

lboxell:
Related to https://github.com/gslab-econ/template/pull/45


gslab_scons/tests/test_configuration_tests.py
```diff
@@ -0,0 +1,394 @@
...
+
+        # Default correct test
+        configuration_tests.check_python_packages('3.0.2', ['yaml', 'os'])
+        mock_convert_packages.assert_called_with(['yaml', 'os'])
+        mock_import.assert_called()
+        mock_version.assert_called_with('gslab_python')
+
+        # Incorrect gslab_python version
+        with self.assertRaises(ex_classes.PrerequisiteError):
+            configuration_tests.check_python_packages('3.0.3', ['yaml', 'os'])
```
M-R-Sullivan: Maybe we could also have a test that ensure the lexicographical ordering issue I mentioned [here]() doesn't cause unintended behaviour?

gslab_scons/tests/test_configuration_tests.py
```diff
@@ -0,0 +1,394 @@
...
+    @mock.patch('gslab_scons.configuration_tests.check_r_packages')
+    def test_check_r(self, mock_check_r_packages, mock_is_in_path):
+        # No R executable
+        mock_is_in_path.side_effect = lambda x: None
+        with self.assertRaises(ex_classes.PrerequisiteError):
+            configuration_tests.check_r()
+        mock_check_r_packages.assert_not_called()
+
+        # Test default arguments 
+        mock_is_in_path.side_effect = lambda x: True
```
M-R-Sullivan: It could be a bit clearer to write `lambda x: (x == 'R')`, but I don't mind the current implementation.

gslab_scons/tests/test_configuration_tests.py
```diff
@@ -0,0 +1,394 @@
...
+        mock_check_r_packages.assert_called_with("yaml")
+
+
+    @mock.patch('gslab_scons.configuration_tests.subprocess.check_output')
+    @mock.patch('gslab_scons.configuration_tests.convert_packages_argument')
+    def test_check_r_packages(self, mock_convert_packages, mock_check_output):
+        # Setup
+        def output_side_effect(*args, **kwargs):
+            cmd_line_call = args[0]
+            print cmd_line_call
```
M-R-Sullivan: Did you intend to leave this print here?

gslab_scons/tests/test_configuration_tests.py
```diff
@@ -0,0 +1,394 @@
...
+
+    @mock.patch('gslab_scons.configuration_tests.is_in_path')
+    def test_check_lyx(self, mock_is_in_path):
+        # No lyx executable
+        mock_is_in_path.side_effect = lambda x: None
+        with self.assertRaises(ex_classes.PrerequisiteError):
+            configuration_tests.check_lyx()
+
+        # Test default 
+        mock_is_in_path.side_effect = lambda x: True
```
M-R-Sullivan: See my comment on line 109.

gslab_scons/configuration_tests.py
```diff
@@ -0,0 +1,180 @@
...
+            raise PrerequisiteError('''
+                              Either Git LFS is not installed 
+                              or your Git LFS settings need to be updated. 
+                              Please install Git LFS or run 
+                              'git lfs install --force' if prompted above.''')
+
+def check_and_expand_cache_path(cache):
+    error_message = " Cache directory, '%s', is not created. " % cache + \
+                    "Please manually create before running.\n\t\t" + \
+                    "    Or fix the path in user-config.yaml.\n"
```
lboxell:
```
Enter cache value: test
PrerequisiteError:  Cache directory, 'test', is not created. Please manually create before running.
		    Or fix the path in user-config.yaml.
:
```
I did it to align with the first line. I'm not super attached to it.

gslab_scons/tests/test_configuration_tests.py
lboxell: Nope.

gslab_scons/tests/test_configuration_tests.py
```diff
@@ -0,0 +1,394 @@
...
+
+        # Failures
+
+        # No yaml value and no default value in path
+        mock_stata_exec.side_effect = lambda x: None
+        with self.assertRaises(ex_classes.PrerequisiteError):
+            configuration_tests.check_stata()
+
+        # Bad package
+        with self.assertRaises(ex_classes.PrerequisiteError):
```
M-R-Sullivan: I think you should undo line 257 (`mock_stata_exec.side_effect = lambda x: None`) before making this check so we know the error is due to the bad package.

gslab_scons/configuration_tests.py
```diff
@@ -0,0 +1,180 @@
...
+    for pkg in packages:
+        try:
+            importlib.import_module(pkg)
+        except ImportError:
+            missing_packages.append(pkg)
+
+    if len(missing_packages) > 0:
+        raise PrerequisiteError('Missing %s module(s)' % missing_packages)
+
+    if pkg_resources.get_distribution('gslab_tools').version < gslab_python_version:
```
lboxell: Yup. Fixed.

lboxell:
Preliminary comments above should be implemented in template and gslab_python.

gslab_scons/tests/test_configuration_tests.py
```diff
@@ -0,0 +1,394 @@
...
+        # Yaml file exists, is not corrupt, and has key.
+        make_yaml('key: value')
+        self.assertEqual(configuration_tests.load_yaml_value('yaml.yaml', 'key'), 'value')
+
+        # Yaml file exists, is not corrupt, and doesn't have key. User enters value for key.
+        make_yaml('bad_key: bad_value')
+        self.assertEqual(configuration_tests.load_yaml_value('yaml.yaml', 'key'), 'value')
+        
+        # But now yaml has a correct key and doesn't require user input.
+        mock_raw_input.side_effect = lambda x: "bad_value"
```
M-R-Sullivan: Consider using `return_value`?

gslab_scons/tests/test_configuration_tests.py
```diff
@@ -0,0 +1,394 @@
...
+        os.remove('yaml.yaml')
+        self.assertEqual(configuration_tests.load_yaml_value('yaml.yaml', 'key'), 'value')
+
+        # Yaml file does not exist. User enters none value to create. 
+        os.remove('yaml.yaml')
+        mock_raw_input.side_effect = raw_input_side_effect2
+        self.assertEqual(configuration_tests.load_yaml_value('yaml.yaml', 'stata_executable'), None)
+
+        # Yaml file now exists with none value.
+        self.assertEqual(configuration_tests.load_yaml_value('yaml.yaml', 'stata_executable'), None)
```
M-R-Sullivan: This would be true even if the file didn't exist (i.e. this line is the same as the previous assert statement). Maybe we could check if the yaml file exists directly?

gslab_scons/tests/test_configuration_tests.py
```diff
@@ -0,0 +1,394 @@
...
+            command = args[0].split("which ")[0]
+            if not re.search("statamp", command):
+                raise subprocess.CalledProcessError(1, args[0])  
+
+            with open('stata.log', 'wb') as f:
+                if not re.search('yaml', args[0]):
+                    package = args[0].split("which ")[1]
+                    f.write("command %s not found" % package)
+
+        def is_unix_side_effect_true():
```
M-R-Sullivan: Consider `return_value` (I'll stop commenting this now and leave it to you to decide where it applies).

gslab_scons/tests/test_configuration_tests.py
```diff
@@ -0,0 +1,394 @@
...
+            return False
+
+        mock_check_output.side_effect     = check_output_side_effect
+        mock_package_argument.side_effect = lambda x: ['yaml']
+        mock_is_unix.side_effect          = is_unix_side_effect_true
+
+        with mock.patch('gslab_scons.configuration_tests.sys.platform', 'win32'):
+            # Good tests
+
+            # Unix
```
M-R-Sullivan: It's slightly confusing that the Unix check occurs while `sys.platform` is patched to `win32`. The current implementation is correct, so I'll leave it to you re: whether to make a change.

gslab_scons/tests/test_configuration_tests.py
```diff
@@ -0,0 +1,394 @@
...
+            configuration_tests.check_stata_packages("statamp", "yaml")
+
+            # Windows
+            mock_is_unix.side_effect = is_unix_side_effect_false
+            configuration_tests.check_stata_packages("statamp", "yaml")
+
+            # No packages
```
M-R-Sullivan: Maybe add assertions to confirm that the function is making the check we expect it to?

gslab_scons/tests/test_configuration_tests.py
```diff
@@ -0,0 +1,394 @@
...
+                configuration_tests.check_stata_packages("statamp", "bad_package")
+            
+            with self.assertRaises(ex_classes.PrerequisiteError):
+                # Bad executable
+                mock_package_argument.side_effect = lambda x: ['yaml']
+                configuration_tests.check_stata_packages("bad_executable", "yaml")
+
+        # More bad tests
+        with self.assertRaises(ex_classes.PrerequisiteError): 
+            mock.patch('gslab_scons.configuration_tests.sys.platform', 'Unknown')
```
M-R-Sullivan: Can you patch like this? I thought you have to use `with mock.patch()` or something similar.

gslab_scons/tests/test_configuration_tests.py
```diff
@@ -0,0 +1,394 @@
...
+
+        mock_version.side_effect          = version_side_effect
+        mock_convert_packages.side_effect = lambda x: ['yaml', 'os']
+        mock_import.side_effect           = lambda x: None
+
+        # Default correct test
+        configuration_tests.check_python_packages('3.0.2', ['yaml', 'os'])
+        mock_convert_packages.assert_called_with(['yaml', 'os'])
+        mock_import.assert_called()
+        mock_version.assert_called_with('gslab_python')
```
M-R-Sullivan: Should this be `gslab_tools`? (See line 31 in `configuration_tests.py`).

gslab_scons/tests/test_configuration_tests.py
lboxell: I decided to keep as is, since we already use side-effect for this mock. I think it would be confusing to mix side_effect and return_value.

gslab_scons/tests/test_configuration_tests.py
lboxell: Added a comment.

gslab_scons/tests/test_configuration_tests.py
lboxell: I think so. But I edited things for consistency.

lboxell:
@M-R-Sullivan, I implemented your comments above.

gslab_scons/tests/test_configuration_tests.py
M-R-Sullivan: OK!

gslab_scons/tests/test_configuration_tests.py
M-R-Sullivan: OK!

M-R-Sullivan:
On hold.

gslab_scons/builders/build_matlab.py
```diff
@@ -53,10 +54,8 @@ def build_matlab(target, source, env):
...
     if misc.is_unix():
         options = '-nosplash -nodesktop'
-    elif sys.platform == 'win32':
-        options = '-nosplash -minimize -wait'
     else:
-        raise Exception("Unknown OS")
+        options = '-nosplash -minimize -wait'
 
     # Run matlab on source file
     shutil.copy(source_file, 'source.m')
```
arosenbe: I'm sure there's a reason, but can you explain why we need this copy step?

gslab_scons/builders/build_matlab.py
```diff
@@ -53,10 +54,8 @@ def build_matlab(target, source, env):
     # Matlab options
     if misc.is_unix():
         options = '-nosplash -nodesktop'
```
M-R-Sullivan: Indeed! I'll fix it.

gslab_scons/builders/build_matlab.py
M-R-Sullivan: @arosenbe - I don't know offhand why we copy the script, and I think it would be cleaner not to do so if making a change makes neither running nor logging more complicated. I'll look into this and report back.

M-R-Sullivan:
@arosenbe - thanks for looking this over! My responses to the points in your review comment:
- I fixed the spacing
- If we think that giving users more control over the commands generated by the builders would be helpful, we should certainly consider adding that functionality. Do you have any particular examples of when this control would be useful? If so, we can add your point to the outline.

gslab_scons/builders/build_matlab.py
M-R-Sullivan: There is no obvious way to run a script in another directory, so I'd support keeping the current implementation for now.

arosenbe: I see. That's fine, and the workaround is clean anyway.

arosenbe:
This branch looks good to me. Is there any reason you'd want to keep it open? If not, let's merge.

M-R-Sullivan:
@arosenbe - I updated the version number to four and reflected this in the template. Merging now!

Issue #69, opened 2017-07-03
Update tablefill builder with 4.0 logging machinery


stanfordquan:
The builder for tablefill doesn't use any of the nice machinery we built for the other builders. We thought it may be a bit complicated to update them but then we realize a hackish way to get around it without having to touch legacy code (i.e. `tablefill.py`).

Basically we'll write another wrapper script for `tablefill.py` with a `__main__` function, and then calls it from subprocess as you would Matlab/R/Stata the likes. Then the logging machinery works normally and you can produce a `Sconscript.log`

stanfordquan:
See PR #70

PR #70, opened 2017-07-03
Pull request for #69: Update tablefill builder with 4.0 logging machinery

stanfordquan:
I found a neater way to do the logging. Turns out `tablefill` has a built-in output return that either prints a happy message when everything works out or a traceback if there's an error. I add code to catch that error and `raise ExecCallError(message)` and have the timestamp matches our conventions. 

Before there was a traceback on Terminal output but no Sconscript log. Sconstruct.log afterwards look like [this](https://gist.github.com/stanfordquan/5ad2b41f96867370fc12fd2f66d6fba6). The Sconscript.log file follows our current logging procedure (timestamps).

### DIFF
--- a/gslab_scons/builders/build_tables.py
+++ b/gslab_scons/builders/build_tables.py
@@ -1,6 +1,7 @@
 import gslab_scons.misc as misc
 from gslab_fill import tablefill
-
+from gslab_scons import log_timestamp
+from gslab_scons._exception_classes import ExecCallError
 
 def build_tables(target, source, env):
     '''Build a SCons target by filling a table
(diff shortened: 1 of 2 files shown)


gslab_scons/builders/build_tables.py
```diff
@@ -23,16 +24,40 @@ def build_tables(target, source, env):
...
+    log_file = target_dir + '/sconscript.log'
     misc.check_code_extension(target_file, '.lyx')
     
-    tablefill(input    = ' '.join([str(i) for i in source[1:]]), 
-              template = source_file, 
-              output   = target_file)
+    # Command call
+    command = """tablefill(input = %s, 
+                 template = %s, 
+                 output   = %s)""" % (input_string, source_file, target_file)
```
arosenbe: We want the alignment to be correct for the console (not for the written code).

gslab_scons/builders/build_tables.py
```diff
@@ -23,16 +24,40 @@ def build_tables(target, source, env):
...
+    output  = tablefill(input    = input_string, 
+                        template = source_file, 
+                        output   = target_file)
+
+    if "error" in str.lower(output): # if tablefill.py returns an error
```
arosenbe: Will this always work? For example, 

```python
raise Exception
'''
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
Exception
'''
```

gslab_scons/builders/build_tables.py
```diff
@@ -23,16 +24,40 @@ def build_tables(target, source, env):
...
+                 template = %s, 
+                 output   = %s)""" % (input_string, source_file, target_file)
+    output  = tablefill(input    = input_string, 
+                        template = source_file, 
+                        output   = target_file)
+
+    if "error" in str.lower(output): # if tablefill.py returns an error            
+        message = misc.command_error_msg("tablefill.py", command)
+        with open(log_file, 'wb') as f:
+            f.write(output)
```
arosenbe: Seems like we're violating DRY with the log writing.

gslab_scons/builders/build_tables.py
```diff
@@ -23,16 +24,40 @@ def build_tables(target, source, env):
...
     misc.check_code_extension(source_file, '.lyx')
 
+    # Set up input string (list of data tables)
+    input_string = ' '.join([str(i) for i in source[1:]])
+
     # Set up target file (filled table)
     target_file = str(target[0])
+    target_dir  = misc.get_directory(target_file)    
+    log_file = target_dir + '/sconscript.log'
     misc.check_code_extension(target_file, '.lyx')
```
arosenbe: The log file construction should be after the target_file check (swap the two lines)

gslab_scons/builders/build_tables.py
stanfordquan: Ah good catch, I was going off the "Error Found" in `tablefill.py` which is not printed. I'll switch it to "traceback" (which we know will be there because `tablefill.py` returns `traceback.formatexc()`

gslab_scons/builders/build_tables.py
stanfordquan: Not sure why (since we were in the business of creating file names in the previous two lines) but will change

stanfordquan:
Thanks @arosenbe . We agreed offline that all edits were made.

Issue #68, opened 2017-07-02
Remove Google Drive as default drive for `release.py` 


stanfordquan:
From the last meeting we agree that we should allow user to specify a release directory that is not Google Drive. My sense is that we can do that by changing

`release.py` line 35

```
    USER = os.environ['USER']
    if branch == 'master':
        name   = repo
        branch = ''
    else:
        name = "%s-%s" % (repo, branch)
    local_release = '/Users/%s/Google Drive/release/%s/' % (USER, name)
    local_release = local_release + version + '/'
```

where we move `USER` and `local_release` to `user_config.yaml`.

@lboxell, I think you're the most familiar with this code out of current lab members and are least likely to unintentionally break something while working with it. Do you want to make this change? I can do it as well. Thanks!

stanfordquan:
PR #72

PR #72, opened 2017-07-11
Pull request for #68: Remove Google Drive from default release

stanfordquan:
@lboxell , please review this PR and its companion piece over at `template`. I put three things into this pull requests:

1) I removed Google Drive hard coding in `gslab_python` `release.py`. We now release to a folder that you can specify in `user_config.yaml`.

2) I moved `load_yaml_values` from `configuration_test.py` to `misc.py`. I think that makes a lot more sense since loading from `yaml` is something builders and scripts in `gslab_python` needs to do, not just that `configuration_test.py` script.

3) I moved the `state_of_repo` function to use `atexit` instead of the dependency trick. This is a significant bug that I realized. The `release.py` function that MRS wrote depends on being able to run `scons .` Running `scons .` broke the earlier implementation, since the target is now `.` which means that `.` depends on `state_of_repo.log` depends on `.` (since `state_of_repo.log` is in the top directory `.`). Also `atexit` is what we use to run the logging, so I think we should continue using that.

### DIFF
--- a/gslab_scons/_release_tools.py
+++ b/gslab_scons/_release_tools.py
@@ -10,12 +10,12 @@ import subprocess
 
 from _exception_classes import ReleaseError
 
-
 def release(vers, org, repo,
             DriveReleaseFiles = [],  
             local_release     = '',  
             target_commitish  = '', 
-            zip_release       = True):
+            zip_release       = True,
+            token             = None):
     '''Publish a release
 
     Parameters
@@ -31,8 +31,9 @@ def release(vers, org, repo,
     '''
     # Check the argument types
 
-
-    token    = getpass.getpass("Enter a GitHub token and then press enter: ") 
+    if token is None:
+        token = getpass.getpass("Enter a GitHub token and then press enter: ") 
+    
     tag_name = vers
     
     releases_path = 'https://%s:@api.github.com/repos/%s/%s/releases' \
@@ -51,7 +52,16 @@ def release(vers, org, repo,
     json_dump = re.sub('"FALSE"', 'false', json_dump)
     posting = session.post(releases_path, data = json_dump)
     # Check that the GitHub release was successful
-    posting.raise_for_status()
+    try:
+        posting.raise_for_status()
+    except requests.exceptions.HTTPError:
+        message  = "We could not post the following json to the releases path \n" 
+        message  = message + ("https://YOURTOKEN:@api.github.com/repos/%s/%s/releases \n" % (org, repo))
+        message  = message + "The json looks like this:"
+        print(message)
+        for (k,v) in payload.items():
+            print(" '%s' : '%s' " % (k,v))
+        raise requests.exceptions.HTTPError
 
     # Release to drive
     if bool(DriveReleaseFiles):
@@ -285,6 +295,6 @@ def extract_dot_git(path = '.git'):
 
     # Next, find the branch's name
     branch_info = open('%s/HEAD' % path, 'rU').readlines()
-    branch = re.findall('ref: refs/heads/([\w-]+)', branch_info[0])[0]
+    branch = re.findall('ref: refs/heads/(.+)\\n', branch_info[0])[0]
 
     return repo, organisation, branch
--- a/gslab_scons/_release_tools.py
+++ b/gslab_scons/_release_tools.py
@@ -11,7 +11,7 @@ import subprocess
 from _exception_classes import ReleaseError
 
 
-def release(vers, org, repo, 
+def release(vers, org, repo,
             DriveReleaseFiles = [],  
             local_release     = '',  
             target_commitish  = '', 
@@ -23,7 +23,7 @@ def release(vers, org, repo,
     env: an SCons environment object
     vers: the version of the release
     DriveReleaseFiles a optional list of files to be included in a
-        release to Google Drive.
+        release to drive (e.g. DropBox or Google Drive).
     local_release: The path of the release directory on the user's computer.
     org: The GtHub organisaton hosting the repository specified by `repo`.
     repo: The name of the GitHub repository from which the user is making
@@ -53,7 +53,7 @@ def release(vers, org, repo,
     # Check that the GitHub release was successful
     posting.raise_for_status()
 
-    # Release to Google Drive
+    # Release to drive
     if bool(DriveReleaseFiles):
         # Delay
         time.sleep(1)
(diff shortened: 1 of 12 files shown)


stanfordquan:
Amazingly enough in the process of fixing release I have broken `mode=cache`.

image

stanfordquan:
I fixed two errors, one I introduced and one was already in the `regex`, no kidding.

Removing `NoCache` for `state_of_repo.log` fixes the `mode=cache` `IndexError`.

I also fixed a regex error that breaks the reading of the GitHub branch from `user_config.yaml`. The regex basically breaks if there's a `#` or non-alphanumeric character in the branch. I fixed it so that it reads until newline instead (i.e. the whole branch).

gslab_scons/_release_tools.py
```diff
@@ -10,29 +10,30 @@
...
+            token             = None):
     '''Publish a release
 
     Parameters
     ----------
     env: an SCons environment object
     vers: the version of the release
     DriveReleaseFiles a optional list of files to be included in a
-        release to Google Drive.
+        release to drive (e.g. DropBox or Google Drive).
```
lboxell: Do we really need the 'DriveReleaseFiles' argument anymore? It seems that everything in the /release/ folder should be sent to 'local_release'. Right?

gslab_scons/_release_tools.py
```diff
@@ -10,29 +10,30 @@
...
     local_release: The path of the release directory on the user's computer.
     org: The GtHub organisaton hosting the repository specified by `repo`.
     repo: The name of the GitHub repository from which the user is making
         the release.
     '''
     # Check the argument types
 
-
-    token    = getpass.getpass("Enter a GitHub token and then press enter: ") 
+    if token is None:
```
lboxell: I'm not sure this is best practice for passwords/tokens (i.e., having a user type it/enter on command line without hiding it).

stanfordquan: I have thought about the tradeoffs here. It's pretty convenient if you're testing the release. Also we don't really force the user to enter token into `user_config`,  the yaml input would be:

```
github_token_optional: bananabananabanana
```

and if the argument is not there, this logical gate would kick into the normal `getpass.getpass`.

If you think this is so bad practice that it can't be here (even though it's optional) I'm happy to remove.

lboxell: Maybe change to 'github_token', but otherwise, I like that.

gslab_scons/_release_tools.py
stanfordquan: I think that's true from ours perspective, but that doesn't mean we should code that into `_release_tools`. That decision is handled in `release.py`.

stanfordquan: The idea is that someone should be able to use the tool in `release.py` and specify what they want to release. I added an option to `release.py` to add your own file list, otherwise as default we crawl through `/release` and release everything in there.

stanfordquan:
@lboxell please see my responses above. Thanks!

lboxell:
@stanfordquan,

 - I got the following:
```
siepr-mg-ra-imac-1:template leviboxell$ python -m gslab_scons.release version=1.23.3TEST
Enter release value: 
```
which doesn't seem to be very clear. I'm assuming its talking about the local release folder? Can we make these messages more informative? I think dropping the `value` and changing `cache --> cache_directory` and `release --> release_directory` would do the trick. If you disagree we can discuss.

 - We should update the readme template with the above edits under `##### What do I put in 'user-config.yaml'?`

 - We should use the `check_and_expand_cache_path` function on the release path also.

stanfordquan:
Those are very good suggestions, will implement, thanks!

stanfordquan:
I implemented the first two points. I did not do the third one because the way the current paths are coded,

1. the function creates the whole path (`release/template-branch/version`)

2. function forces there to be a superdirectory called `/release/` which is kinda useful for when you're releasing to share drives

3. we don't want to ask the user to make the whole path manually, as is the case in `check_and_expand_path`.

lboxell:
@stanfordquan,

 - It looks like `value` is still present [here](https://github.com/gslab-econ/gslab_python/pull/72/files/a2559ef2d0b43100bf3033c7cb6e1e4444c062e8..e0f6f4b7ef8b0d1af55b131bc6871701d226f2b1#diff-74ed9966e6411ce4518f7174f9b45be0R252) from my first point.

 - It seems like we should replace the directory checking code in the release function with the `check_and_expand_path` function. The `check_and_expand_path` function is for situations where the user enters `~/Dropbox/release/` instead of  `/Users/Example_User/Dropbox/release/` (we had issues with the `~` command before) OR if `/Users/Example_User/Dropbox/release/` doesn't exist yet locally (which avoids duplicate folders being populated on Drive or Dropbox). Beyond that, I think we should just assume the user knows where he wants to put his files (and not force him to have `/release/` in the path name).

stanfordquan:
@lboxell I see what you meant in the first point now. So what you want is for there to be no prompt for the Github token? I'm not sure what the implication of that is and why that's preferred. You're already keeping the token in a plain text `user-config` file , might as well have the added utility of a prompt for an optional input. I can just have the function breaks if it doesn't find the value for the Github token if you prefer that .

stanfordquan:
@lboxell I removed the prompt for the GitHub token and also added `check_and_expand_path` to the release path. The forced `/release/` in the release path is left over from when the code was first written. I removed the restriction and the relevant test.

gslab_scons/misc.py
```diff
@@ -276,11 +273,14 @@ def load_yaml_value(path, key):
         else:
             return yaml_contents[key]
     except:
-        with open(path, 'ab') as f:        
-            val = str(raw_input(prompt % key))
-            if re.sub('"', '', re.sub('\'', '', val.lower())) == "none":
-                val = None
-            f.write('%s: %s\n' % (key, val))
+        if key == "github_token_optional":
```
lboxell: We want `github_token` here right? Or is the readme wrong?

gslab_scons/misc.py
```diff
@@ -245,9 +245,6 @@ def load_yaml_value(path, key):
 
     if key == "stata_executable":
         prompt = "Enter %s value or None to search for defaults: "
-    elif key == "github_token":
-        prompt = "(Optional) Enter %s to be stored in user_config.yaml.\n" + \
-                 "Github token can also be entered without echoing later:" 
     else:
         prompt = "Enter %s value: "
```
lboxell: I was actually referring to putting `Enter %s:` here instead of `Enter %s value:`. The `value` part seems odd to me.

gslab_scons/misc.py
stanfordquan: that's correct, mb

gslab_scons/misc.py
stanfordquan: Ah I see, that's fine with me. I added the prompt for the github_token back in.

stanfordquan:
Fixed those comments!

stanfordquan:
Hallelujah, the errors are from the lines setting `sys.tracebacklimit = 0`.

lboxell:
Looks good to me. The only thing to consider is using getpass for entering the token or add this to the outline.

stanfordquan:
Last commit does what you asked for, @lboxell.

Issue #75, opened 2017-07-19
Print traceback for python builder


arosenbe:
The current Python Builder will raise a custom error and message when a python process exits on error, but it doesn't print the traceback. The goal of this task is to get the Python Builder to print both our custom error and the traceback when a python process exits on error.

stanfordquan:
@arosenbe this was the `tablefill` implementation I was referring to.

```
#! /usr/bin/env python

import os
import argparse
import types
import re
import traceback
import tablefill_info
from decimal import Decimal, ROUND_HALF_UP


def tablefill(**kwargs):
    try:
        args = parse_arguments(kwargs)
        tables = parse_tables(args)
        lyx_text = insert_tables(args, tables)
        lyx_text = insert_warning(args, lyx_text)
        write_to_lyx(args, lyx_text)
        exitmessage = args['template'] + ' filled successfully by tablefill'
        print exitmessage
        return exitmessage	
    except:
        print 'Error Found'
        exitmessage = traceback.format_exc()
        print exitmessage
        return exitmessage
```

arosenbe:
Under PR by @stanfordquan at #76.

PR #76, opened 2017-07-19
Pull request for #75: Print traceback for python builder

arosenbe:
@stanfordquan, can you please review my work from #75? Let me know if you'd like the error message to be printed in another manner. I went with what I thought would be most useful for me, but I don't have strong opinions.


stanfordquan:
Neat, @arosenbe , that's the implementation we want.

I caught a bug in the original logging mechanism. If the `sconscript.log` is empty (because the script fails and there's no timestamp and that the script produces no output on its own), then the logging function breaks due to a `readline` error. I added a try/except that will take care of that.

Issue #77, opened 2017-07-20
Replace <*> with <source file name>  in <sconscript_*.log> 


stanfordquan:
Suppose we have 

` source/analysis/dofile.do`
` source/analysis/mfile.m`

writing to `build/analysis/`

Then they will overwrite `sconscript.log` under the current implementation.

We want them to write to

` build/analysis/sconscript_dofile_do.log`
` build/analysis/sconscript_mfile_m.log` 

respectively.

This should also change the `log.py` particularly in the `os.walk` step to add a wildcard for whatever follows `sconscript` in `sconscript.log`

stanfordquan:
PR lives in #80

PR #80, opened 2017-07-21
Pull request for #77: Replace <*> with <source file name> in <sconscript_*.log>

stanfordquan:
@arosenbe I added a step in builders for the user to specify a bit in the sconscript_*.log as mentioned in the title. The user specify an argument in `env.BuildSomething` as `log_ext = *`. It should be fairly obvious what happens. There's a PR in `template` as a companion to this.

I would appreciate it if you can help out updating the unittests since you're the Mock expert.

Thanks!

### DIFF
--- a/gslab_scons/builders/build_lyx.py
+++ b/gslab_scons/builders/build_lyx.py
@@ -39,7 +39,11 @@ def build_lyx(target, source, env):
 
     misc.check_code_extension(source_file, 'lyx')
     newpdf   = source_file.replace('.lyx','.pdf')
-    log_file = target_dir + '/sconscript.log'
+    try:
+        log_ext = '_%s' % env['log_ext']
+    except:
+        log_ext = ''
+    log_file    = target_dir + ('/sconscript%s.log' % log_ext)
     
     # System call
     try:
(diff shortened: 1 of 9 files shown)


arosenbe:
#82 has some fixes for the tests of build_table. We'll hold off here until that goes through to avoid redundant work.

stanfordquan:
Tests run fully now, @arosenbe. Thanks!

gslab_scons/builders/build_lyx.py
```diff
@@ -39,7 +39,11 @@ def build_lyx(target, source, env):
 
     misc.check_code_extension(source_file, 'lyx')
     newpdf   = source_file.replace('.lyx','.pdf')
-    log_file = target_dir + '/sconscript.log'
+    try:
+        log_ext = '_%s' % env['log_ext']
+    except:
```
stanfordquan: Sure, but that broke a bunch of unittests that feed weird objects as `Env`

gslab_scons/builders/build_lyx.py
```diff
@@ -39,7 +39,11 @@ def build_lyx(target, source, env):
...
+        log_ext = ''
+    log_file    = target_dir + ('/sconscript%s.log' % log_ext)
```
stanfordquan: Done

stanfordquan:
@arosenbe I responded to tyour PR but that broke a number of unittests that feed weird objects as `Env`, so `env['log_ext']` raises different kinds of errors instead of KeyError and break the code.

stanfordquan:
@arosenbe I fixed the unittests.

Issue #73, opened 2017-07-14
Allow tablefill to fill .tex documents


lboxell:
Currently, tablefill isn't compatible with .tex. The goal of this task is to make it so.

lboxell:
Closing task.

Pull request: https://github.com/gslab-econ/gslab_python/pull/82

PR #82, opened 2017-07-24
Pull request for #73: Allows tablefill to use tex documents

lboxell:
@stanfordquan,

This should be a pretty short PR, so I think we can slot it into 4.0 release.


stanfordquan:
I have reviewed and approved.

Issue #89, opened 2017-07-31
debrief error on new repo


lboxell:
@stanfordquan,

Do the `issue_size_warnings` in `scons_debrief` only get printed to the `state_of_repo.log`? If so, I think we might want to remove them. I think there are two issues with the current setup:
 (1) I think [this line](https://github.com/gslab-econ/gslab_python/blob/master/gslab_scons/size_warning.py#L63) causes issues on a repo that has been cloned, but doesn't have any version history yet.
 (2) I get errors if I don't have `/release/` as a folder. Though I don't understand why not having `/raw/` doesn't cause similar issues (see [here](https://github.com/gslab-econ/gslab_python/blob/master/gslab_scons/size_warning.py#L10)). 

Perhaps I'm missing something. 

I'm getting: 
```
scons: building terminated because of errors.
fatal: your current branch 'master' does not have any commits yet
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/Users/leviboxell/anaconda/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/leviboxell/anaconda/lib/python2.7/site-packages/gslab_scons/misc.py", line 26, in scons_debrief
    issue_size_warnings(look_in, file_MB_limit, total_MB_limit)
  File "/Users/leviboxell/anaconda/lib/python2.7/site-packages/gslab_scons/size_warning.py", line 17, in issue_size_warnings
    versioned = create_size_dictionary(look_in)
  File "/Users/leviboxell/anaconda/lib/python2.7/site-packages/gslab_scons/size_warning.py", line 121, in create_size_dictionary
    raise ReleaseError("The path argument does not specify an "
ReleaseError: The path argument does not specify an existing directory.
Error in sys.exitfunc:
Traceback (most recent call last):
  File "/Users/leviboxell/anaconda/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/leviboxell/anaconda/lib/python2.7/site-packages/gslab_scons/misc.py", line 26, in scons_debrief
    issue_size_warnings(look_in, file_MB_limit, total_MB_limit)
  File "/Users/leviboxell/anaconda/lib/python2.7/site-packages/gslab_scons/size_warning.py", line 17, in issue_size_warnings
    versioned = create_size_dictionary(look_in)
[...]

PR #90, opened 2017-08-01
Pull request for #89: Fixes issues on new repo

lboxell:
@stanfordquan, I implemented fixes to the two issues. I can't remember whether we want to merge bug fixes directly to master or not. In this case, it seems like we'd want to (and raise to 4.0.1), but I'm open to other options. I attached what the state of repo looks like on a repo with no git history, which I think is good enough for our purposes, but let me know if you think otherwise. 
[state_of_repo.txt](https://github.com/gslab-econ/gslab_python/files/1192126/state_of_repo.txt)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# GSLab Python Library Collection 4.0.0
+# GSLab Python Library Collection 4.0.1
 
 Overview
 --------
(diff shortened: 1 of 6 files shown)


stanfordquan:
Looks good, @lboxell . I'd say pull into master and make bug-fix release 4.0.1.

Issue #92, opened 2017-08-09
Change executable-finding behavior in Stata builder


stanfordquan:
Code uses default path to a StataMP executable for mac or windows.

User is allowed to specify to the executable in config_user.yaml.

If neither path works and Stata dependencies are enabled, raise informative error.

stanfordquan:
Pull request #104.

stanfordquan:
Found a pre-existing error in `configuration_tests.py`, where this get printed if you enter nothing into the stata_executable prompt (which is bad!). You don't get it if you enter None or your actual executable.

```
scons: Reading SConscript files ...
Enter stata_executable or None to search for defaults:
PrerequisiteError: Stata command, '', failed.
		   Please supply a correct stata_executable value in user-config.yaml.
:
```

stanfordquan:
Turns out I already fixed that... I was on `master` instead of `4.1.0` in `gslab_python`.

PR #104, opened 2017-08-22
Pull request for #92: Change executable-finding behavior in Stata builder

stanfordquan:
- Remove "is executable in path" searching in Stata builder. Now the builder defaults to `stata-mp` on Mac and `StataMP-64.exe` on Windows if `user-config.yaml` does not have a `stata_executable` key, or specify as blank, or specify as "None". 

- The builder sends a message if the builder fails while using one of the default options.

- As the template is set up, the first Stata check is in the `check_stata_packages`  function run in `configuration_test` (since we run stata `which` from the command line). This means that you will never get a builder error for Stata if the configuration test includes running Stata and then fail. This is something to bring up with MG.

### DIFF
--- a/gslab_scons/builders/build_stata.py
+++ b/gslab_scons/builders/build_stata.py
@@ -44,10 +44,9 @@ def build_stata(target, source, env):
     
     # Set up command line arguments
     cl_arg = misc.command_line_args(env)
-
-    executable       = misc.get_stata_executable(env)
+    executable = misc.get_stata_executable(env)
     command_skeleton = misc.get_stata_command(executable)
-
+    
     try:
         command = command_skeleton % (source_file, cl_arg)
         subprocess.check_output(command, 
@@ -55,6 +54,8 @@ def build_stata(target, source, env):
                                 shell  = True)
     except subprocess.CalledProcessError:
         message = misc.command_error_msg("Stata", command)
+        if  env['stata_executable'] not in [None, 'None', '']:
+            message = message + 'Maybe try specifying the Stata executable in config_user.yaml?'
         raise ExecCallError(message)
 
     shutil.move(loc_log, log_file)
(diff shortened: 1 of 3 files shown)


gslab_scons/builders/build_stata.py
```diff
@@ -44,17 +44,18 @@ def build_stata(target, source, env):
...
-
+    
     try:
         command = command_skeleton % (source_file, cl_arg)
         subprocess.check_output(command, 
                                 stderr = subprocess.STDOUT,
                                 shell  = True)
     except subprocess.CalledProcessError:
         message = misc.command_error_msg("Stata", command)
+        if  env['stata_executable'] not in [None, 'None', '']:
```
Shun-Yang: I'm a little confused on why we need this `if` condition.

stanfordquan: Good catch, that should be an `if in` not an `if not in`

Shun-Yang: I see. What happens when user enter a wrong stata-executable and then stata fails? Do we want to raise a comment saying that "maybe you should recheck your stata-executable"?

stanfordquan: If the user enters the wrong executable, what we said is the stock "we run this command and it failed," and echo the command used in terminal. 

If the subprocess failed and the user did not specify a Stata executable (hence the `if`) we print the additional warning.

stanfordquan:
@Shun-Yang is there more to be said on this PR? Thanks!

Shun-Yang:
@stanfordquan , I don't have other comments. Just to confirm, per our noodle discussion, we will change the current behavior on `configuration_test` in `template`?

stanfordquan:
@Shun-Yang that's correct. I prefer that bullet point to be dealt with in gslab-econ/template#63. Let me know if you disagree.

Shun-Yang:
@stanfordquan , agree, clear to merge.

Issue #93, opened 2017-08-09
Move top-level logs to release/


stanfordquan:
Put state_of_repo.log, sconstruct.log, and whatever we use to log externals in release. 

Should be the case that git status after a run of scons only shows assets that changed. This should be done in conjunction with `template` top-level reorganization.

If moving the sconstruct.log to release is taking a long time to figure out, reconsider.

Shun-Yang:
@stanfordquan Please refer to gslab-econ/template#2. I tried with some attempts to move sconstruct.log and state_of_repo.log to release/ directory but didn't succeed. The issue seems to be that the python command will be executed before all scons command run. I don't think there is a good and natural way to do this. We probably want to reconsider on this.

Shun-Yang:
Scratch the last comment. `atexit` works to change the ordering of the python calls.

stanfordquan:
Glad to help! Putting two other suggestions on record (you don't have to implement these, just keeping notes)

- Potentially fold `log_move` into `log_end` instead of extra line in `SConstruct`

- Use `env.Install()` and regex for `scons*.log` files

Shun-Yang:
Closing the task. Pull request is in #107 and parallel pull request is in gslab-econ/template#75. 
Summary: change the `end_log` functionality to move logs to top level directory. Make the same changes in template.

PR #107, opened 2017-08-23
Pull request for #93: add moving top level log to end_log 

Shun-Yang:
@stanfordquan, see #93. The parallel pull is gslab-econ/template#75. Should be quick.


stanfordquan:
I'll run this on template and commit.

Issue #95, opened 2017-08-09
Builders should check whether specified targets exist post-run


stanfordquan:
Determine if we raise an error or warning.
Implement the decision.

Shun-Yang:
My proposal is to raise an error. I can't see a scenario where we are okay with some targets not generated (in that case, we don't need to specify those targets in the first place). Someone [here](https://stackoverflow.com/questions/45096365/fail-if-target-was-not-generated) seems to try to do the same thing. Let me know if you see this differently @stanfordquan @yuchuan2016.

stanfordquan:
I agree that raising the error is the way to go.

Should it be raised after a build step or after the complete build (`atexit`?) We're loading a lot of stuff into `atexit`, so we may want to abstract that out to an `atexit` function suite.

yuchuan2016:
I agree. Shall we add this in `scons_debrief`?

Shun-Yang:
@stanfordquan  I would propose to raise an error after a build step instead of the complete build, as there is no need to continue building if an error arises.  @yuchuan2016 , What is `scons_debrief` and where I can find info on that?

yuchuan2016:
Sorry for the confusion. It's a function in gslab_scons.misc. But if we want to raise error after each step, then it's of no use.

Shun-Yang:
Closing the issue. Pull request is at #110. 
Summary: I added a functionality to check if prespecified target files exit or not after run. If not, an specific error will arise for the user.

PR #110, opened 2017-08-24
Pull request for #95: Check target files after each building step

Shun-Yang:
@yuchuan2016 , please see #95. I have test this a bit on template but please test it more and try to break it if you can. Thanks!


Shun-Yang:
@yuchuan2016 , I think my last commit has addressed everything. Is it okay for you to address "adding the target in the side_effect" in #109?

yuchuan2016:
@Shun-Yang , OK!

Issue #111, opened 2017-08-24
Fix release.py for the new top level structure


Shun-Yang:
We noticed that `release.py` does not work properly given our current top level structure. Need to fix it sometime.

Shun-Yang:
Closing the issue. Pull request is at #113.

PR #113, opened 2017-08-24
Pull request for #111: Fix release.py

Shun-Yang:
@yuchuan2016 , this should be straightforward!

### DIFF
--- a/gslab_scons/README.md
+++ b/gslab_scons/README.md
@@ -1,7 +1,7 @@
 ## Notes on release.py
 
 Make a release from an SCons directory by running the following
-from the command line within the directory of interest:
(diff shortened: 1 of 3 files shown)


yuchuan2016:
@Shun-Yang , I'm not sure whether we want to simply change it to `../.git`. `gslab_python` is not only used in `template` and does not necessarily have the same structure. I would propose we first search for `.git` in the current directory, and if nothing is found, we search the upper level directory. What do you think?

Shun-Yang:
@yuchuan2016 , good point. I have implemented your suggestion in the last commit.

yuchuan2016:
@Shun-Yang , great! It works well.

Issue #97, opened 2017-08-09
Write Anything builder


stanfordquan:
Use SCons's env.Command functionality to allow arbitrary build steps with our logging functionality.

stanfordquan:
[Another instance of searching for Scons help returning the gslab-econ/ra-manual.](https://www.google.com/search?q=does+scons+rerun+if+%22env%22+change+environment&oq=does+scons+rerun+if+%22env%22+change+environment&gs_l=psy-ab.3..33i21k1.5357.6538.0.6617.12.12.0.0.0.0.148.1063.5j5.10.0....0...1.1.64.psy-ab..2.9.915...33i160k1.XAWxObcFAxc). We're approaching gslab singularity.

stanfordquan:
@gentzkow 

Hi Matt. The anything builder that I came up with use a `do` argument in the environment passed to the builder to read what the code it should run in the terminal is. However this is bad because when this `do` argument change, Scons does not recognize that the action has changed.

I see three options:

- Force user to write down this `do` argument in a script (does not have to be `sh`, can be anything). The builder reads this script as a source and execute it verbatim. In the below example, this script would just be one line `statamp -e source/analysis/analysis.do`.

- Abandon this `Anything` builder and instead have a `sh` builder and  a `cmd` builder that run shell/batch code. This will be in the same spirit as our Python builder, with code extension check, so forth.

- Leave as is.

I think the first one is ideal, since the user can conceivably have things they can't call with `sh` or `batch`. Please let me know if you disagree.

```
env.BuildAnything(target = ['#%s/plot.pdf'    % build,
                            '#%s/table.txt'    % build],
                  source = ['#%s/analysis.do' % source],
                  do = 'statamp -e source/analysis/analysis.do')
```

gentzkow:
Hmm. I would have thought we want the builder to replicate as closely as possible what the Command builder does, with the added logging. Can we not keep the syntax identical to the Command builder?

stanfordquan:
@gentzkow,

I read into the `env.Command` a bit more --- what it does is that it generates a custom builder, with the custom `action` at the heart of the custom builder. I followed that approach, the syntax is now 

```python
gs.build_anything(target = ['#%s/data.txt'    % build],
                  source = ['#%s/create_data.do' % source],
                  action = 'stata-mp -e source/anything/create_data.do',
                  env    = env) # only showing mandatory arguments
```

The `env` argument has to be there because it's not automatically passed
[...]

PR #114, opened 2017-08-25
Pull request for #97: Anything builder

stanfordquan:
Please see #97 for discussion. There's a branch on the template that will demonstrate a use case. I will update the template FAQ after this pull request and before the issue closes.

### DIFF
--- a/gslab_scons/builders/build_anything.py
+++ b/gslab_scons/builders/build_anything.py
@@ -1,0 +1,107 @@
+import os
+import subprocess
+import gslab_scons.misc as misc
+import warnings
+import re
+from gslab_scons import log_timestamp
+from gslab_scons._exception_classes import ExecCallError
+
+def build_anything(target, source, env):
+    ''' 
+    Anything builder. A wrapper that will run any code in the terminal
+    using gslab_scons builder logging mechanism.
+    
+    The user must specify a "do" and a log_file" arguments when 
+    calling the builder.
+    
+    The command called in the user's terminal will be
+        [do] > [log_file]
+    
+    Parameters
+    target: string or list 
+        The target(s) of the SCons command. The log ends up in the 
+        directory of the first specified target.
+    source: string or list
+        The source(s) of the SCons command. 
+    env: SCons construction environment, see SCons user guide 7.2
+        env must contain the following arguments:
+        do: string
+            The command to be run in terminal. Must not contain a redirection 
+            operator i.e. > or >>
+        log_file: string
+            The log file where output will be redirected. Must be a full path
+            from the directory's root, ending with `sconscript*.log` where *
+            can be anything. 
+        [OPTIONAL]
+        origin_log_file: string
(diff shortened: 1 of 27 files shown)


stanfordquan:
@Shun-Yang these two links may help. I borrowed the structure from [`env.Command`](http://scons.org/doc/HTML/scons-api/SCons.Environment-pysrc.html#Base.Command) and the [`Builder`](http://scons.org/doc/HTML/scons-api/SCons.Builder-module.html) module help page explains a bit more about what a Builder/Node is.

gslab_scons/builders/build_anything.py
```diff
@@ -0,0 +1,104 @@
...
+        target_dir  = misc.get_directory(target_file)
+        try:
+            log_ext = '_%s' % env['log_ext']
+        except KeyError:
+            log_ext = ''
+        log_file = os.path.join(target_dir, ('sconscript%s.log' % log_ext))
+
+        # System call
+        try:
+            command = '%s > %s' % (action, log_file)
```
Shun-Yang: What about `command = '%s > %s  2>&1' % (action, log_file)` to also log `stderr`?

gslab_scons/builders/build_anything.py
```diff
@@ -0,0 +1,104 @@
...
+        log_ext: string
+            Instead of logging to `sconscript.log` in the target dir, 
+            the builder will log to `sconscript_<log_ext>.log`.
+        origin_log_file: string
+            Sometimes, your command may produce a log file in an undesirable location.
+            Specifying the that location in this argument leads the builder to append
+            the content of origin_log_file to log_file and delete origin_log_file.            
+            The builder will crash if this file doesn't exist at the end of the command.
+        warning: Boolean
+            Turns off warnings if warning == False.
```
Shun-Yang: `warning = False`?

stanfordquan:
1) Done, thanks!

2) Added newline.

3) Yes we can make it say `build_anything`, but then the action generator needs a different name. How about `_build_anything` for the generator, and then `build_anything` for the main event?

4) Basically, no. `Stata` has this funky thing of producing a log at the root directory with the name of a dofile, I was trying to prevent that sort of behavior (which is, so far, only necessary in the Stata case). In the other builders you almost always catch all your logs with the redirection operator `>`.  

5) Agreed.

stanfordquan:
@Shun-Yang I made another change in the `_build_anything` `warning` syntax. Basically because it's dealt with within the generator and not the builder, it's passed as a keyword rather than an `env` element.

I made the warning slightly prettier.

image

Shun-Yang:
@stanfordquan , I think the directory `anything_test` is accidentally merged to template 4.1.0 branch. Should we delete it?

stanfordquan:
Yes @Shun-Yang . I think I was on the wrong branch when I made that commit.

stanfordquan:
Done as of7dfbad4.

Issue #94, opened 2017-08-09
Address concerns about builder logs and SCons targeting


stanfordquan:
Builder logs are versioned but not targets. So pulling from the cache on a clean clone makes it look like all the logs were deleted.
Determine if we specify logs as targets.
Is there a one-to-one mapping between build steps and logs? Do we allow this mapping to be broken? Allow folder to be excluded in builder log searches.
What is the default name scheme for logs, and how flexible is it? Do we allow users to overwrite/customize just the name or the path as well (probably yes to both)?

stanfordquan:
1) Builder logs are located where the output of a folder is, so whether they are being versioned depends on whether that folder is versioned. 
    - If the folder is versioned, then you never need to pull anything from cache, so this is not a concern. 
    - If the folder is *not* versioned, I don't see a significant cost of not having a `SConscript.log` when you pull from `cache`. The build obtained from `cache` has to be consistent with the code currently in the directory (required by SCons to have the same build signature). 
    - The issue with setting builder logs as targets and then caching them is that you break 1-1 mapping between source code and target, since builder logs contain:
        - Timestamps (this is our own doing)
        - Local machine information (e.g. Stata version, local directory names. This is unavoidable)

So I recommend against targeting/versioning builder logs.  

2) There can and should be a one-to-one mapping between build steps and logs, but it is not enforced. We specify a build step as followed:

```python
env.BuildR(target  = ['#%s/data_r.txt'    % build],
           source  = ['#%s/create_data.r' % source],
           log_ext =  'create_data_r')
```

This logs the build step to `<build>/sconscript_create_data_r.log`. Absent this argument, the build step is logged to `<build>/sconscript.log`. 1-1 mapping is ensured when the user is conscious about which logs they are writing to which output folder, but not otherwise. 

So moving forward, there are two things to be done:

- Allowing the output log to be in a "free location" that is not the target directory, by way of a `log_location` argument to be passed the same way `log_ext` is passed.

- When the `sconstruct.log` maker search the directory for recently-created `sconscript.log` to be appended, allow certain folders to be excluded, by way of an argument when calling `end_log` at the end of the SConstruct.


@yuchuan2016 @Shun-Yang can either of you confirm my impressions and comment on my plan above before I ping PIs? Thanks!

stanfordquan:
Note: improve IOError when  ` # move top level logs to /release/ directory.`

yuchuan2016:
[...]

PR #115, opened 2017-08-28
Pull request for #94: allow folders to be excluded from search

stanfordquan:
@yuchuan2016 this should be fairly simple.

I added an argument in `collect_builder_logs` called `excluded_dirs` (which is passed by the user when calling `log.end_log` at `atexit` in `template`. This argument is a str or list of str of directories to be excluded when we search for sconscript.log files to collect in the `sconstruct.log`.

Let me know if there's any question! Thanks

### DIFF
--- a/gslab_scons/log.py
+++ b/gslab_scons/log.py
@@ -74,20 +74,30 @@ def log_timestamp(start_time, end_time, filename = 'sconstruct.log'):
     return None
 
 
-def collect_builder_logs(parent_dir):
+def collect_builder_logs(parent_dir, excluded_dirs = []):
     ''' Recursively return dictionary of files named sconscript*.log 
         in parent_dir and nested directories.
         Also return timestamp from those sconscript.log 
-        (snippet from SO 3964681)'''
+        (snippet from SO 3964681)
+
+        excluded_dirs (str or list of str):
+            list of directories to be excluded from the search
+        '''
     builder_log_collect = {}
 
     # Store paths to logs in a list, found from platform-specific command line tool 
     rel_parent_dir = os.path.relpath(parent_dir)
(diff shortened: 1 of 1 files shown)


gslab_scons/log.py
```diff
@@ -74,20 +74,30 @@ def log_timestamp(start_time, end_time, filename = 'sconstruct.log'):
...
+        '''
     builder_log_collect = {}
 
     # Store paths to logs in a list, found from platform-specific command line tool 
     rel_parent_dir = os.path.relpath(parent_dir)
     log_name = 'sconscript*.log'
+    excluded_dirs = misc.make_list_if_string(excluded_dirs)
     if misc.is_unix():
-        command = 'find %s -name "%s"' % (rel_parent_dir, log_name)
+        command = 'find %s -name "%s -type f"' % (rel_parent_dir, log_name)
```
yuchuan2016: I think this should be `'find %s -name "%s" -type f'`.

yuchuan2016:
@stanfordquan , good work! I have no other comment except the above one. I have another releated comment. The current search for `sconscript*.log` is in all sub-directories. Then if a log file in `build` is also sent to `release`, it will be appended to `sconstruct.log` twice. Shall we worry about this?

stanfordquan:
That's a good point @yuchuan2016. I didn't plan on having duplicated logs around a directory, so I haven't thought about this.

In principle I'm ok with this. This "appending" mechanism is mostly for user readability while working on code / debugging, so I'm fine if there's some redundant information. Otherwise I'm not really sure about how we detect which logs are copied and which logs are created new.

yuchuan2016:
@stanfordquan , great. I'm also fine with this.

PR #117, opened 2017-08-29
Pull Request for template#63: Make current dependencies optional

yuchuan2016:
This is in conjunction with https://github.com/gslab-econ/template/pull/78.


yuchuan2016:
We no longer need to change `gslab_python` for gslab-econ/template#78. Pull request abandoned.

PR #120, opened 2017-08-31
Pull request for gslab-econ/template#60: Provenance

stanfordquan:


gslab_scons/provenance.py
```diff
@@ -0,0 +1,195 @@
...
+                          start_path = '.',
+                          verbose = False):
+    
+    files = misc.make_list_if_string(external_provenance)
+    if files == [] or find_for_me == True:
+        pattern = 'provenance*.log'
+        files += misc.finder(start_path, pattern, excluded_dirs)
+    files = sorted(files)
+    
+
```
arosenbe: Delete extra line

gslab_scons/provenance.py
```diff
@@ -0,0 +1,195 @@
...
+                          verbose = False):
+    
+    files = misc.make_list_if_string(external_provenance)
+    if files == [] or find_for_me == True:
+        pattern = 'provenance*.log'
+        files += misc.finder(start_path, pattern, excluded_dirs)
+    files = sorted(files)
+    
+    
+    with open(root_provenance, 'a') as root_f:
```
arosenbe: Open in `'ab'`.

gslab_scons/provenance.py
```diff
@@ -0,0 +1,195 @@
...
+            warning_message = '\nThe file %s does not appear to be a GSLab provenance file.\n' % provenance
+            warn = False
+            
+            if os.path.normpath(root_provenance) != os.path.normpath(provenance):
+                root_f.write('*** Subdirectory provenance: %s\n' % os.path.abspath(provenance))
+                with open(provenance, 'rU') as f:
+                    
+                    try:
+                        content = f.readlines()
+                        if content[0].rstrip() != '*** GSLab directory provenance ***': warn = True
```
arosenbe: String should only be written in one place. Define once and reference with a variable.

gslab_scons/provenance.py
```diff
@@ -0,0 +1,195 @@
...
+                    except IndexError:
+                        warn = True
+
+                    if warn == True:
+                        warnings.warn(warning_message)
+                        root_f.write(warning_message)
+                    root_f.write(''.join(content))
+                root_f.write('\n\n')
+
+    return None
```
arosenbe: New line at end.

gslab_scons/provenance.py
arosenbe: Why rstrip instead of strip?

gslab_scons/provenance.py
```diff
@@ -0,0 +1,195 @@
...
+    with open(root_provenance, 'a') as root_f:
+        for provenance in files:
+            if verbose == True:
+                print os.path.abspath(provenance)
+            
+            warning_message = '\nThe file %s does not appear to be a GSLab provenance file.\n' % provenance
+            warn = False
+            
+            if os.path.normpath(root_provenance) != os.path.normpath(provenance):
+                root_f.write('*** Subdirectory provenance: %s\n' % os.path.abspath(provenance))
```
arosenbe: Subdirectory -> External dependency?

gslab_scons/provenance.py
```diff
@@ -0,0 +1,195 @@
...
+                with open(provenance, 'rU') as f:
+                    
+                    try:
+                        content = f.readlines()
```
arosenbe: Move readlines below file opening. Move try/except block to same indentation as `with`.

gslab_scons/provenance.py
arosenbe: Move within the for loop over files (just above `if warn == True` and at same indentation level as opening of each sub-provenance). Collect all lines that write to the provenance file in one place.

stanfordquan:
See example of release [here](https://github.com/gslab-econ/template/releases/tag/test_aug30_c)

gslab_scons/provenance.py
```diff
@@ -100,11 +101,13 @@ def scan(start_path, include_details, include_checksum, detail_limit, file_detai
     return total_size, num_files, last_mtime, file_details, dirs
 
 
-def write_heading(start_path, provenance_path):
+def write_heading(start_path, provenance_path, github_release = None):
```
arosenbe: No need to set the default value for github_release in the sub-function since it's already set globally in make_provenance.

gslab_scons/provenance.py
```diff
@@ -100,11 +101,13 @@ def scan(start_path, include_details, include_checksum, detail_limit, file_detai
...
     '''
     Write standard heading for provenance: what and for where it is. 
     '''
     out = '*** GSLab directory provenance ***\ndirectory: %s\n' % os.path.abspath(start_path) 
+    if github_release != None:
+        out += '\n*** GitHub release: %s \n' % github_release
```
arosenbe: Breaks pattern since the release information appears on the same line as the section header. Consider breaking the text over two lines.

gslab_scons/provenance.py
```diff
@@ -129,8 +132,13 @@ def write_readme(readme_path, provenance_path):
     '''
     Writes readme to provenance.
     '''
-    with open(readme_path, 'rU') as f:
-        out = '%s' % f.read()
+    try:
+        with open(readme_path, 'rU') as f:
+            out = '%s' % f.read()
+    except IOError:
+        raise IOError(('%s does not exist.\n If you are in release mode,' % readme_path) +
```
arosenbe: Error message isn't true. The same exception will be raised if you try to read a directory.

gslab_scons/provenance.py
```diff
@@ -129,8 +132,13 @@ def write_readme(readme_path, provenance_path):
...
+                      'please specify the command line argument param `readme=<README_PATH>` .' )
```
arosenbe: Drop space before closing period?

gslab_scons/provenance.py
```diff
@@ -129,8 +132,13 @@ def write_readme(readme_path, provenance_path):
     '''
     Writes readme to provenance.
     '''
-    with open(readme_path, 'rU') as f:
-        out = '%s' % f.read()
+    try:
+        with open(readme_path, 'rU') as f:
+            out = '%s' % f.read()
+    except IOError:
```
arosenbe: Think we could shorten these error messages to 80 characters per line?

gslab_scons/provenance.py
arosenbe: We can use `sys.argv[0]` to determine the name of the script that's being executed. We should check if that value ends in `release.py` and only print the release-related stuff if it does.

gslab_scons/provenance.py
```diff
@@ -9,11 +9,14 @@
 def make_provenance(start_path, readme_path, provenance_path, 
                     include_details = True,
                     include_checksum = True,
-                    detail_limit = 500,
-                    external_provenance = [],
-                    find_for_me = False,
-                    excluded_dirs = [],
-                    verbose = False):
+                    detail_limit = 500,         # max number of files to output details for
```
arosenbe:
```python
# max number of files to calculate and output details for
```

gslab_scons/provenance.py
```diff
@@ -9,11 +9,14 @@
...
+                    external_provenance = [],   # list of external_provenance. If empty, function automatically looks up provenance files.
```
arosenbe:
```python
# list of external provenance files to append. If empty, automatically look for provenance files.
```

gslab_scons/provenance.py
```diff
@@ -9,11 +9,14 @@
...
+                    find_for_me = False,        # automatically looks for provenance files regardless of external_provenance
+                    excluded_dirs = [],         # exclude these directories from automatic-look-up
```
arosenbe:
```python
# exclude these directories from automatic provenance look-up
```

gslab_scons/release.py
```diff
@@ -49,6 +50,18 @@ def main(user_yaml = 'config_user.yaml', release_files = []):
...
+    try:
+        readme = next(arg for arg in sys.argv if re.search("^readme=", arg))
+    except:
+        readme = "readme=./README.md"
+
+    readme = re.sub('^readme=', '', readme)
+    provenance_release_note = '%s, version %s' % (name, version)
+    make_provenance(start_path = '.',
+                    readme_path = readme,
+                    provenance_path = './release/provenance.log',
```
arosenbe: I think we should allow a user running release.py to control some of the default options in make_provenance.

gslab_scons/provenance.py
stanfordquan: I think that's overly complicating things. This kind of message suffice, you aren't constantly running `provenance.py` anyway.

gslab_scons/release.py
stanfordquan: As of the moment we hard code the release directory to be `./release` so I think this is fine.

gslab_scons/provenance.py
arosenbe: Won't we be directly running provenance.py for raw data on Dropbox?

stanfordquan: Yup.

gslab_scons/provenance.py
```diff
@@ -0,0 +1,195 @@
+import os
+import sys
+import datetime
+import warnings
+import scandir # pip install scandir
+import mmh3 # pip install mmh3
+import gslab_scons.misc as misc
+
+def make_provenance(start_path, readme_path, provenance_path,
```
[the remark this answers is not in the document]
stanfordquan: We already call it provenance_path everywhere else, I don't think it's that confusing

gslab_scons/provenance.py
```diff
@@ -0,0 +1,195 @@
...
+    '''
+    file_details = 'path|bytes|most recent modification time'
+    if include_checksum:
+        file_details = '%s|MurmurHash3' % file_details
+    
+    return [file_details]
+
+
+def scan_wrapper(start_path, include_details, include_checksum, detail_limit, file_details,
+                 verbose = False):
```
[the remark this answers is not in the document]
stanfordquan: I prefer having default option in helper functions that can conceivably be called outside of our upper level functions

gslab_scons/provenance.py
stanfordquan: No idea what you mean here

gslab_scons/provenance.py
stanfordquan: Per discussion we're fine with subdirectory

gslab_scons/provenance.py
stanfordquan: Going to keep it as a default option in write_heading and append_sub_provenance called sig

gslab_scons/provenance.py
stanfordquan: I prefer having default option for helper functions

gslab_scons/provenance.py
stanfordquan: It's not really a section header as writing `directory: <stuff>`

gslab_scons/provenance.py
stanfordquan: GSLab standard is 100 cols no?

gslab_scons/provenance.py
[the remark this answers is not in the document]
arosenbe: Can you implement that everywhere then?

gslab_scons/provenance.py
arosenbe: Can I just come over and show you?

gslab_scons/provenance.py
arosenbe: If it's not a section heading, then it shouldn't be formatted as a section heading (`***`). Maybe just format it like the `directory` entry?

gslab_scons/provenance.py
arosenbe: Yes, but pep is 80, and I think it makes things easier to read. Leave it if you want.

gslab_scons/release.py
arosenbe: I meant more about controlling things like `find_for_me` and `detail_limit`.

gslab_scons/provenance.py
arosenbe: Makes sense. Can you implement everywhere then.

gslab_scons/provenance.py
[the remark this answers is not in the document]
arosenbe: I meant changing it everywhere.

gslab_scons/release.py
stanfordquan: With the exception of `excluded_dirs` (which you have to activate by writing a wrapper for `__main__` rather than calling using `python -m` the rest can now be specified with command line args.

stanfordquan:
@arosenbe your comments are addressed in the above commits.

gslab_scons/provenance.py
```diff
@@ -101,11 +136,13 @@ def scan(start_path, include_details, include_checksum, detail_limit, file_detai
     return total_size, num_files, last_mtime, file_details, dirs
 
 
-def write_heading(start_path, provenance_path, github_release = None):
+def write_heading(start_path, provenance_path, 
+                  github_release = None,
+                  sig = '*** GSLab directory provenance ***'):
```
arosenbe: I'd love it if we defined a sig variable in one location and then passed it to all the functions that use it as a default argument. Hard coding the default in two places is not robust because the two defaults can get out of sync.

gslab_scons/release.py
arosenbe: Make sure we mention the command line args somewhere.

gslab_scons/provenance.py
```diff
@@ -2,42 +2,56 @@
...
+                                                                  include_details  = include_details,
+                                                                  file_details     = file_details,
+                                                                  include_checksum = include_checksum,
+                                                                  detail_limit     = detail_limit, 
+                                                                  verbose          = verbose)
+                                                     
+
+        write_heading(start_path, provenance_path, github_release)
+        write_directory_info(provenance_path, total_size, num_files, last_mtime)
+        write_readme(readme_path, provenance_path)
```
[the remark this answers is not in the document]
stanfordquan: I'm fine with keeping them modular. Having another helper that calls five other helpers is not appealing to me. Neither is concatenating the five helpers into one giant helper.

gslab_scons/provenance.py
arosenbe: Per discussion, we're formatting the `directory` entry like `Github release`.

gslab_scons/release.py
```diff
@@ -61,11 +61,11 @@ def main(user_yaml = 'config_user.yaml',
 
     # Check provenance options
```
arosenbe: I'd like users to be able to set verbose and (if possible) excluded_dirs as well. It seems silly to have these options that make a user's life easier, if the user can't access them in the standard use case.

gslab_scons/provenance.py
```diff
@@ -2,42 +2,56 @@
...
+        if include_details:
+            write_detailed_info(provenance_path, details)
+        
+        write_ending(provenance_path)
+        append_sub_provenance(provenance_path, external_provenance, excluded_dirs,
+                              find_for_me, start_path, verbose)
+    except Exception as e:
+        print('make_provenance.py failed.')
+        print(e)
```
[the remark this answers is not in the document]
stanfordquan: True, I replaced this with `raise e`.

gslab_scons/_release_tools.py
```diff
@@ -15,7 +15,8 @@ def release(vers, org, repo,
             local_release     = '',  
             target_commitish  = '', 
             zip_release       = True,
-            github_token      = None):
+            github_token      = None,
+            provenance_path   = './release/provenance.log'):
     '''Publish a release
```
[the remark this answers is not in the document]
stanfordquan: Which one is out of date?

gslab_scons/provenance.py
```diff
@@ -2,42 +2,56 @@
...
-
-    total_size, num_files, last_mtime, details = scan_wrapper(
-        start_path, include_details, include_checksum, detail_limit, file_details, verbose)
+    try:
+        file_details = determine_file_details(include_details = include_details,
+                                              include_checksum = include_checksum)
+
+        total_size, num_files, last_mtime, details = scan_wrapper(start_path       = start_path,
+                                                                  include_details  = include_details,
+                                                                  file_details     = file_details,
```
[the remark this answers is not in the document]
stanfordquan: It is the last non-optional argument

gslab_scons/provenance.py
```diff
@@ -2,42 +2,56 @@
...
+                                                                  include_checksum = include_checksum,
+                                                                  detail_limit     = detail_limit, 
+                                                                  verbose          = verbose)
+                                                     
+
```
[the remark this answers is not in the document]
stanfordquan: Done

gslab_scons/release.py
stanfordquan: `excluded_dirs` can't be set from command line since it's a list. `verbose` is fine.

gslab_scons/provenance.py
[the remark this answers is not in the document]
arosenbe: Can we make the print statement above read more nicely. It's not clear that people who are releasing their work will know what make_provenance.py is. The script is also called provenance.py.

gslab_scons/provenance.py
```diff
@@ -41,11 +40,11 @@ def make_provenance(start_path, readme_path, provenance_path,
                               find_for_me, start_path, verbose)
     except Exception as e:
         print('make_provenance.py failed.')
```
arosenbe: See https://github.com/gslab-econ/gslab_python/pull/120#discussion_r136410044. This line isn't nice to read, and it points too a non-existant script.

stanfordquan: meant to write `make_provenance()`

gslab_scons/release.py
arosenbe: It could be passed as a string, which we convert into a list, but yeah that might be a little much. I'm also not positive of the use case for that argument.

gslab_scons/provenance.py
arosenbe: Is a user of `release.py` guaranteed to know what `make_provenance()` is/does?

stanfordquan: `release.py` has its own README and guide which we will update. I treat `release.py` as an advanced functionality, not something new users will touch much. That's how we currently treat it in template as well

arosenbe: Great point! We should document these optional CL args in that readme. Want to do it as part of this task, or just note it in gslab-econ/template#82. I'd vote for this task.

Issue #122, opened 2017-09-04
Delete old local gslab_python versions upon both installation methods


stanfordquan:
At the moment a developer/user can install gslab_python in two ways:

1) Clone gslab_python and run `python setup.py install clean`

2) Clone template (as of 4.1.0) and run `python config/config_python.py`. 

The problem is that 1) writes the `egg` version into `site_package` and 2) writes both the `egg` and the subfolders (`gslab_blahblah`) into `site_package`. What this leads to is the following behavior:

a) Install gslab_python using 2)

b) Install different version of gslab_python using 1)

Now the versions of gslab_python on your computer are out of sync.

Our solution: remove *both* the `egg` and the subfolders upon installation of gslab_python using either method. Implemented by a search and destroy function in `setup.py`, then test whether `python config/config_python.py` properly runs that script when installing `gslab_python`.

arosenbe:
We want to warn the user, explain what's going on, and ask for permission before search and destroy. Also make sure this works on Windows (no idea about redirection in that setting).

arosenbe:
Under PR at #123. What a fun number.

arosenbe:
PR passed at #123. 

`setup.py` now checks for any gslab-stamped packages (or gencat) in the default package installation location and deletes whatever it finds.

PR #123, opened 2017-09-05
Pull request for #122: Delete old local gslab_python versions upon both installation methods

arosenbe:
@stanfordquan, can you please PR my work from #122. The idea is to look through the current default python package installation location and destroy every gslab package there. We don't ask, and we don't tell. This's basically the same behavior as pip, just a little quieter. 

I opted for shutil.rmtree over `pip uninstall` because it's sure to catch all versions of GSLab, even if multiple versions exist of they were installed through unorthodox means. 

One thing to note is that this implementation only checks the default package installation location. If the user specifies another location, may God have mercy on her soul...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ if True in is_include_arg:
 else:
     include_arg = None
 
-with open('Users/arosenbe/Desktop/A', 'wb') as f:
+with open('/Users/arosenbe/Desktop/A', 'wb') as f:
     f.write('AAAA')
 
 class TestRepo(build_py):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -17,7 +19,16 @@
 else:
     include_arg = None
 
+# Uninstall old versions of GSLab-Tools
+re_gslab  = re.compile('gslab[-_].', re.IGNORECASE)
+re_gencat = re.compile('gencat')
+package_location = site.getsitepackages()[0]
```
stanfordquan: I don't see an `import site` statement

arosenbe: Good catch ❤️

Issue #99, opened 2017-08-09
Add cache reconfiguration code


stanfordquan:
Can switch between one and two numeral naming.

#### Add our own instructions on configuring the SCons cache to the readme and add code to gslab_scons. See [here](https://github.com/gslab-econ/gslab_python/issues/71)

arosenbe:
I've done a bit of searching, and it seems that the discrepancy in single vs. double indexing of the cache is related to the SCons version. Versions before 2.5.0 use single indexing and no `config` file in the cache, while versions after use double indexing and a `config` file to store that indexing information. [source](https://bitbucket.org/scons/scons/src/rel_2.5.1/src/CHANGES.txt?at=2.5.1&fileviewer=file-view-default#CHANGES.txt-60)

When a version of SCons >= 2.5.0 encounters a cache created by an older version, it recognizes the discrepancy and yells at you. 

When the opposite happens, and an old SCons encounters a cache created by the newer version, it ignores the cached, double-indexed files and populates the cache with freshly built, single-indexed files. Running SCons >= 2.5.0 on this mixed up cache will simply ignore the single-indexed files, and it won't yell at you. 

I think the issue wasn't that SCons switched between single and double indexing. Rather some users had a version of SCons before 2.5.0 and others had version 2.5.0 or later. New users shouldn't encounter this problem because SCons >= 2.5.0 is currently the default for pip, homebrew, anaconda, and [SCons's download page](http://scons.org/pages/download.html).

---

I suggest that we do not package `scons-configure-cache.py`. Rather we do one of the following:
* Package scons-local 2.5.1
* Mention the issue with older versions of SCons (true dependency is SCons >= 2.4.0) and point to SCons's own solution in our FAQ.

arosenbe:
@yuchuan2016, can you look over my suggestion at https://github.com/gslab-econ/gslab_python/issues/99#issuecomment-326145577? I'd lean toward packaging scons-local (which I think you have experience with) but am open to suggestions.

arosenbe:
@lboxell, thought you might be interested in https://github.com/gslab-econ/gslab_python/issues/99#issuecomment-326145577.

lboxell:
@arosenbe,

That seems consistent with what I'd seen before. For some reason, it was happening every so often in the lab still. I guess there could have been an old version of scons floating around on someones desktop/sherlock/etc.

arosenbe:
[...]

Issue #98, opened 2017-08-09
Determine the syntax of R command line call


stanfordquan:
Right now we use R CMD Batch. I'm pretty sure there was a good reason for that (maybe Windows?), so check past threads.

Check Google, R documentation, stack overflow, Tobias.
Make the change in the R builder if necessary.

Shun-Yang:
@stanfordquan , I couldn't find where we have discussed about R command line call in the past threads. Let me know if you know anything or have a pointer to the thread.

stanfordquan:
@Shun-Yang Please see gslab-econ/template#56 (search for R CMD BATCH).

Shun-Yang:
@stanfordquan Thanks! I also found that, but I thought we may have discussed why we use `R CMD BATCH` before. Anyway, here is my deliverable: 
### Options
3 options: `R CMD BATCH`, `littler` and `Rscript`. 

### Comparison

#### General: 

`R CMD BATCH` is outdated as most current R users are using `Rscript` or `little r` for command line call. 
There are few things that are inconvenient with `R CMD BATCH` as listed in the following sub-section. `Rscript` should be able to do everything we currently have with `R CMD BATCH` and is also shipped with R.  `littler` is faster than `Rscript` but needs to be installed and is very hard to run on Windows. Therefore, I would suggest us switch to `Rscript`. 

#### Specifics: 

R CMD BATCH:
- echo the input statements
- can not output to stdout

Rscript:
- does not echo
- output to stdout
- can be used in one-liner (i.e. with no input file)
- can use `shebang` line 

littler:
- all that Rscript does
- can read commands from stdin 
- faster startup time
- load the methods package
- can use `shebang` line 
- **hard to run on Windows**

### Reference

The above summary refers to these three sources: 
- https://stackoverflow.com/questions/21969145/why-or-when-is-rscript-or-littler-better-than-r-cmd-batch
- https://stackoverflow.com/questions/3205302/difference-between-rscript-and-littler
- https://stackoverflow.com/questions/3506007/running-r-code-from-command-line-windows
- https://stackoverflow.com/questions/18306362/run-r-script-from-command-line

Shun-Yang:
@stanfordquan , see above, if you agree, I think I will go ahead and implement `Rscript`.

stanfordquan:
Go for it! Thanks @Shun-Yang

Shun-Yang:
[...]

Issue #96, opened 2017-08-09
Write LaTeX builder


stanfordquan:
Write a LaTeX builder with our logging functionality. See latexmk as something we might want to use.

stanfordquan:
@yuchuan2016 can you please also pull #102 in? You can make comments and I can edit the code. I think it'd be a useful introduction into how you may approach the LaTeX builder.

yuchuan2016:
Pull request created in #105.

PR #125, opened 2017-09-05
4.1.0 Release

stanfordquan:
notes to follow


gslab_scons/misc.py
```diff
@@ -11,39 +11,42 @@
 import _exception_classes
 from size_warning import issue_size_warnings
 
-def scons_debrief(target, env):
-    '''Execute functions after SCons has built all targets'''
+def scons_debrief(args):
```
stanfordquan: This introduce backward incompatibility, since the user has to modify SConstruct. @Shun-Yang I don't have a great grasp on this function and you were the last one to touch it, can you add a quick fix?

stanfordquan: ^ Comment as part of #130 #133

Shun-Yang: We can change `args['log']` back into the function argument as `log = 'state_of_repo.log'` but the reason we moved it is that this looks more compact and natural.

stanfordquan: We had a discussion in the lab just now. Our solution is to provide a user-advisory that you cannot pass `env` and `log` to the function any longer and have to pass `args`, at which point we point to the current SConstruct in `template`. So we're fine with both 1) backward incompatibility and 2) coupled gslab_python/template for this particular problem.

stanfordquan:
# gslab_python 4.1.0 Release Notes

## Features

- We added a LaTeX builder and an "Anything"-builder. (#96 #97)

- We added code to reconfigure the SCons cache (#99)

- We add a target-checker after a builder has run (#95)

- `gslab_scons` features now use a local version of SCons by default (#111 #126)

- We add a feature to produce `state_of_inputs` (#131) 

- `gslab_python` installation now delete old local versions (#122)

## Misc

- We move top-level logs to `release/.` or a user-specified location (#93)

- We allow suffixes to the name of builder logs (#94)

- We updated R syntax (#98) and Stata-executable searching behavior (#92)

arosenbe:
@stanfordquan, I think this is ready to go. Do you want to make the pull?

Issue #131, opened 2017-09-06
Change provenance to state_of_inputs


arosenbe:
We currently create static provenance files at release. We want these to be dynamic and produced at runtime (on `scons`). The idea is to record the state of local copies of external inputs. See https://github.com/gslab-econ/template/issues/86#issuecomment-327526608 for details.

The goal of this task is to implement this change. It will, among other things, purge provenance creation from the release procedure. 

This task subsumes #128.

stanfordquan:
PR in #132

arosenbe:
PR passed at #132. Summary comment
* Destroy all references to provenance.py
* Port walk engine from provenance.py to [record_dir.py](https://github.com/gslab-econ/gslab_python/blob/e8b0ecdb27513768fab83a5978ff975ec27a1306/gslab_scons/record_dir.py) 
* Create record_dir.py to record some characteristics of files in a directory and append the characteristics to a log.

PR #132, opened 2017-09-06
Pull request for #131: Change provenance to state_of_inputs

arosenbe:
@stanfordquan, can you review my work from #131? This is a rethink of the way we use the provenance making tool. The actual engine is pretty much the same. 

Tandem to gslab-econ/template#86 and gslab-econ/template#87


gslab_scons/record_dir.py
```diff
@@ -0,0 +1,113 @@
+import os
+import scandir
+import mmh3
+
+from misc import check_and_expand_path, make_heading
+
+def record_dir(start_path, name, 
+               include_checksum = False,
+               file_limit = 5000, 
+               logpath = 'state_of_external.log'):
```
stanfordquan: `logpath` vs `start_path` kinda throws me off but also it's the kind of syntax I would personally use so who knows, you decide.

gslab_scons/record_dir.py
```diff
@@ -0,0 +1,113 @@
...
+    '''
+    Record relative path, size, and (optionally) checksum of all files within start_path.
+    Relative paths are from start_path.
+    Append info in |-delimited format to logpath below a heading made from start_path.
+    '''
+    check_logpath(logpath)
```
stanfordquan: So where can we possibly get a bad path? I see `logpath` and `start_path`. We should see how we want to deal with this. I don't think throwing an error is a good idea, since it's possible that this subdirectories didn't exist prior to running builders / SCons creating all the necessary subdirectories. Maybe writing "we can't find your directory 🤷‍♀️ " to file?

gslab_scons/record_dir.py
arosenbe: Now `outpath` and `inpath` respectively. Hard to mix those up.

gslab_scons/record_dir.py
arosenbe: We can get bad paths if the user specifies 
* inpath as a directory that doesn't exist
* outpath in a directory that doesn't exist

New implementation does the following
* `inpath`: Check that `inpath` exists. If not, add ', could not find at runtime.' to its name. If `inpath` exists as a file, change `inpath` to the directory and record that we only want information about that file. 
* `outpath`: Check that the dir where we put `outfile` is a directory. If not, try to make it. (If the directory exists as a file, we throw an error.)

arosenbe:
@stanfordquan, can you take a look at my changes?

stanfordquan:
I think the conflict above simply is that you should delete `provenance.py`.

Issue #130, opened 2017-09-05
Check and enforce principle for uncoupling gslab_python from gslab template


stanfordquan:
Our principle is that folder paths in `gslab_python` should have defaults that match the folder structure in `template`, but these should be easily modified. 

This task should double check that this principle is upheld in `gslab_python` codebase

arosenbe:
@stanfordquan, @yuchuan2016, @Shun-Yang 

Anyone want to do this task? I'm probably not optimal since I wasn't here for a lot of the code changes at the start of the sprint. It's also the last item for the milestone before the PIs take another look.

stanfordquan:
Collecting points where we couple directory structure with `gslab_python`.

PR #133, opened 2017-09-11
Pull request for #130: Uncouple directory structure from gslab_python

stanfordquan:
I did a check for where we may have hard-coded template directory structure into gslab_python. 

I listed those points [in this comments](https://github.com/gslab-econ/gslab_python/issues/130#issuecomment-328647545). 

In peer review, please check that the list is correct and complete, and that you agree with my decisions / fixes. Thanks!

### DIFF
--- a/gslab_scons/log.py
+++ b/gslab_scons/log.py
@@ -35,7 +35,7 @@ def end_log(log = 'sconstruct.log', excluded_dirs = []):
         f.write(end_message)
 
     # scan sconstruct.log for start time
-    with open('sconstruct.log', "rU") as f:
+    with open(log, "rU") as f:
         s = f.readline()
         s = s[s.find('{') + 1: s.find('}')]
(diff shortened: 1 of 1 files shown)


gslab_scons/log.py
```diff
@@ -60,8 +60,11 @@ def end_log(log = 'sconstruct.log', excluded_dirs = []):
                 sconstruct.write(sconscript.read())
 
     # move top level logs to /release/ directory.
+    release_dir = './release/'
```
Shun-Yang: Can we make `release_dir = './release/'` as an argument for this function instead of writing this inside the function?

stanfordquan:
Thanks @Shun-Yang. I'm going to keep point #1 as is since I think the name of `state_of_repo.log` is essentially the product of a `gslab_python` functionality rather than something intrinsic about structure of template.

stanfordquan:
I fixed the above comment in `b8d2020`. Jumped the gun a bit on the pull.

PR #138, opened 2017-09-20
Pull request for 4.1.0: Change log exclusion syntax to prune

stanfordquan:
@arosenbe , I made a quick change to the excluding folder from `log.endlog()` syntax. It works better now. Earlier you would have to specify the `./` in the path.

### DIFF
--- a/gslab_scons/misc.py
+++ b/gslab_scons/misc.py
@@ -7,6 +7,7 @@ import subprocess
 import datetime
 import yaml
 import getpass
+import fnmatch
 # Import gslab_scons modules
 import _exception_classes
 from size_warning import issue_size_warnings
@@ -342,9 +343,12 @@ def finder(rel_parent_dir, pattern, excluded_dirs = []):
     '''
 
     if is_unix():
-        command = 'find %s -name "%s" -type f' % (rel_parent_dir, pattern)
-        for x in excluded_dirs: # add in args to exclude folders from search
-            command = '%s -not -path "*%s*"' % (command, os.path.normpath(x)) 
+        command = 'find %s' % (rel_parent_dir)
+        if len(excluded_dirs) > 0:
+            for x in excluded_dirs: # add in args to exclude folders from search
+                command = '%s -path "*%s*" -prune -o ' % (command, os.path.normpath(x))
+        command = '%s -name "%s" -type f' % (command, pattern)
+
     else:
         command = 'dir "%s" /b/s' % os.path.join(rel_parent_dir, pattern)
         for x in excluded_dirs:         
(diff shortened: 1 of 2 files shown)


stanfordquan:
@arosenbe please see this answer https://stackoverflow.com/a/16595367

gslab_scons/misc.py
```diff
@@ -342,9 +343,12 @@ def finder(rel_parent_dir, pattern, excluded_dirs = []):
     '''
 
     if is_unix():
-        command = 'find %s -name "%s" -type f' % (rel_parent_dir, pattern)
-        for x in excluded_dirs: # add in args to exclude folders from search
-            command = '%s -not -path "*%s*"' % (command, os.path.normpath(x)) 
+        command = 'find %s' % (rel_parent_dir)
+        if len(excluded_dirs) > 0:
+            for x in excluded_dirs: # add in args to exclude folders from search
+                command = '%s -path "*%s*" -prune -o ' % (command, os.path.normpath(x))
```
arosenbe: Aren't you missing some `-not` switches and `\(`-type bracketing? I really can't tell.

It might be clearer to construct the base `command` and the `excluded_dirs` separately and then join them together in a third step. This method moves `command` around a lot, making it hard to keep track of.

gslab_scons/misc.py
```diff
@@ -354,6 +358,7 @@ def finder(rel_parent_dir, pattern, excluded_dirs = []):
         out_paths = subprocess.check_output(command, shell = True).replace('\r\n', '\n')
         out_paths = out_paths.split('\n')
         out_paths = filter(bool, map(str.strip, out_paths)) # Strip paths and keep non-empty
+        out_paths = fnmatch.filter(out_paths,pattern)
```
arosenbe: Space after comma.

gslab_scons/misc.py
stanfordquan: I agree it's super confusing (and quite frankly a miracle that it worked that well). Here's a breakdown. Suppose I want to find needle `*sconscript*.log` in haystack `.`, but want to avoid `build/data` and `build/external`.  The following two expressions are almost equivalent:

```
find . -path "*build/data*" -prune -o  -path "*build/external*" -prune -o  -name "*sconscript*.log" -type f
```

```
find . -not \( -path "*build/data*" -prune -o  -path "*build/external*" -prune \)  -name "*sconscript*.log" -type f
```

The first one says: in haystack `.`, prune `build/data` (i.e. if you hit `build/data`, skip and move on) OR prrune `build/external` OR search for needle. 

The second one (this is the one in the SO) says in hay stack `.`, if you did not prune (i.e. didn't hit) `build/data` AND did not prune `build/external` THEN you search for needle. 

I use the first one since it's a bit easier to understand for me.

gslab_scons/misc.py
stanfordquan: Used the Auto-Pep-8-ifier, hopefully that'll make everyone equally displeased.

Issue #140, opened 2017-12-06
Fix order of operations in release


arosenbe:
* Avoid update to `sconstruct.log` during scons dry run

<details>
  <summary>Practical example</summary>

![screen shot 2017-12-06 at 10 02 59 am](https://user-images.githubusercontent.com/19935793/33677178-c2a30e6a-da6c-11e7-9883-33dcea8bfd23.png)

</details>

<p>
 
* Need to drop second [`upload_asset` call](https://github.com/gslab-econ/gslab_python/blob/d0697da5dee9befff2340a62cb47dc72f3c0fb31/gslab_scons/_release_tools.py#L141), which is a holdover from our adventures with `provenance.log`.

<details>
  <summary> Practical example </summary>

![screen shot 2017-12-06 at 10 04 50 am](https://user-images.githubusercontent.com/19935793/33677260-ff214884-da6c-11e7-88d1-db535977b9ea.png)
*but my release still went through because of the [first call](https://github.com/gslab-econ/gslab_python/blob/d09a27720a7f264b24000747c087bb977c33c931/gslab_scons/_release_tools.py#L132)*

</details>

arosenbe:
The commit above drop the second upload_asset call. This fix is a lot easier than the fix for the dry run. Here's the issue. 
* Set up
    * gslab_python isn't allowed to know the path of the sconstruct.log
    * The path of the sconstruct.log is currently set in template's SConstruct. 
    * The release to GH only includes the most recently committed versions
    * The release to "other local location" (e.g. Dropbox) includes the current contents of the directory
* Problem 
    * Checking that all SCons targets are up-to-date executes a dry run. 
    * gslab_scons logs the dry run to a location set in template. 
    * The location in template defaults to release/sconstruct.log. So the default behavior makes a change to committed files
    * Executing the release send the committed (no long most recent) version of the log to GH and the uncommitted (now most recent) version of the log to the local path.

arosenbe:
I went for the fix of executing the `git status` before the `scons --dry-run`. This makes the first attempt at a release successful if everything is in its intended state. I also noted the log changing behavior in the readme.

arosenbe:
Under PR by @stanfordquan at #141.

arosenbe:
PR passed at #141. Summary 
* Switches order of operations in release (check git before scons). 
* Give all loggers their own files.
* Add code to determine if SCons is executing in dry-run mode.
* Toggle loggers off when SCons is executing in dry-run-mode.

PR #141, opened 2017-12-19
Pull request for #140: Fix order of operations in release

arosenbe:
@stanfordquan, can you review my work from #140? I know we discussed the proper way to check that the repository is up-to-date, and implemented what I thought the best option was. Let me know if you want to discuss more.


arosenbe:
Thanks @stanfordquan, 

1. No reason that I can think of. I could make up some excuse about scons always being cheap to rerun when you should be making a release. We also might have included it as a back door to avoid the issue with the sonstruct.log changing. I agree that we should have both or neither. Now that we're changing the order so that it's possible to make an error free release, I'd vote neither. Thoughts?

2. I agree that it's two functions and that it's confusing. I also think it's cheap to split. So I will.

stanfordquan:
Agree with both. 🍜

arosenbe:
@stanfordquan, can you take a look at my change above? I think it should address the comments.

gslab_scons/release.py
```diff
@@ -22,7 +22,13 @@ def main(version = None,
...
         except:
             pass
-    inspect_repo(scons_local_path = scons_local_path)
+
+    # Check if repository is up-to-date and ready for release. Stop if not.
+    # Order matters because SCons check changes logs.
+    elif not _release_tools.git_up_to_date():
+        raise ReleaseError('Git working tree not clean.')
+    if not _release_tools.scons_up_to_date(scons_local_path):
+        raise ReleaseError('SCons targets not up to date.')
```
stanfordquan: Should be `if elif`.

arosenbe: Silly. Fixed.

gslab_scons/_release_tools.py
```diff
@@ -174,79 +169,54 @@ def upload_asset(github_token, org, repo, release_id, file_name,
...
+    '''
+    Execute a git status and check that all changes are committed.
+    '''
+    # If mode = git, check whether the working tree is clean.
+    command = 'git status'
+    log = execute_up_to_date(command)
+    # Determine whether we're in a git repository based on git logging.
+    if check_list_for_regex('Not a git repository', log):
+        raise ReleaseError('git_up_to_date must be run on a git repository.')
+    result = check_list_for_regex('nothing to commit, working tree clean', output)
```
stanfordquan: @arosenbe would you be able to take a closer look at this file? Here the `output` variable isn't defined. You should probably check the `master` version to see how this variable was defined originally.

arosenbe: Ah, sorry again. I changed the name output to log everywhere, except in apparently one location.

arosenbe:
@stanfordquan, can you check again?

stanfordquan:
@arosenbe , I thought we were supposed to avoid this kind of behavior: (maybe I misunderstood)

```
analysis (4.1.0) $ git checkout .
analysis (4.1.0) $ git st
On branch 4.1.0
Your branch is up-to-date with 'origin/4.1.0'.

nothing to commit, working tree clean
analysis (4.1.0) $ scons
scons: done reading SConscript files.
scons: Building targets ...
scons: `build' is up to date.
scons: `release' is up to date.
scons: done building targets.
analysis (4.1.0) $ git st
On branch 4.1.0
Your branch is up-to-date with 'origin/4.1.0'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   release/sconstruct.log
	modified:   release/state_of_repo.log

no changes added to commit (use "git add" and/or "git commit -a")
analysis (4.1.0) $ python -m gslab_scons.release version=test_jan8_5 readme=../README.md
Traceback (most recent call last):
  File "/Users/lequan/anaconda/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/Users/lequan/anaconda/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/lequan/anaconda/lib/python2.7/site-packages/GSLab_Tools-4.1.0-py2.7.egg/gslab_scons/release.py", line 88, in <module>
    main()
  File "/Users/lequan/anaconda/lib/python2.7/site-packages/GSLab_Tools-4.1.0-py2.7.egg/gslab_scons/release.py", line 29, in main
    raise ReleaseError('Git working tree not clean.')
gslab_scons._exception_classes.ReleaseError: Git working tree not clean.
analysis (4.1.0) $
```

arosenbe:
No that's not the behavior we're trying to avoid. You're going into the release with a messy working tree. That's the type of error the git check is supposed to prevent. It should work if you checkout and then release.

arosenbe:
@stanfordquan, on further thought, we could avoid issues like the one above (and skirt the entire reason this issue was created) with a simple change. If our SConstruct was aware of whether SCons was executing a dry run, it could toggle the logging off. That way we'd only update the logs when the targets are updated as well. 

I don't think this would be difficult to implement at all. We would have the template's configuration.py determine if --dry-run was included in the command line arguments, and only start the logging process if it wasn't. Thoughts?

stanfordquan:
This is a point I've always wondered --- should we run the logging on `scons --dry-run`? I've always found it useful (since I'd like to see the change reflected in a `SConstruct.log`when my SCons tree is not clean) but I agree that it's not part of the build process.

stanfordquan:
@arosenbe 

>No that's not the behavior we're trying to avoid. You're going into the release with a messy working tree. That's the type of error the git check is supposed to prevent. It should work if you checkout and then release.

But then you never have a clean Git tree, unless you are supposed to commit `SConstruct.log` and `state_of_repo.log` before you release, and then not running `scons --dry-run` or `scons` prior to running `release.py`? I was not aware that this is the intended workflow.

arosenbe:
Hmm. That's the workflow I always followed. If your working tree isn't clean then you're guaranteed to have different logs on GH and in the local release (but I guess that's the case anyway because of the dry run in release.py). 

I don't think I've ever looked the log after a dry run, since it prints to the shell. But if you think logging dry runs is a desirable feature, then I'm happy not to turn it off. 

Another alternative would be to ignore the `sconstruct.log`, `state_of_repo.log`, and `state_of_external.log` when doing the git check before releasing. I don't have a specific issue with this, only that it seems like a recipe for disaster. 

I think we're in a quadrilemma if we want all these features
1. Execute dry run before release
2. Dry runs follow standard logging process
3. Require clean git working tree for release
4. Don't ignore standard logs when checking git working tree

arosenbe:
Per discussion (by phone, what tech!) we're going to extricate ourselves from the [quadrilemma](https://github.com/gslab-econ/gslab_python/pull/141#issuecomment-356662868) by cutting 2. We'll edit the template to turn off logging during dry runs.

arosenbe:
Opened gslab-econ/template#92 to implement. Will png back here when that's done.

arosenbe:
Template task now in PR at gslab-econ/template#93.

gslab_scons/log.py
```diff
@@ -7,11 +7,13 @@
 import gslab_scons.misc as misc
 
 
-def start_log(mode, vers, log = 'sconstruct.log'):
+def start_log(mode, vers, cl_args_list = sys.argv, log = 'sconstruct.log'):
```
stanfordquan: Have you tested this usage outside of the template? I wonder how `sys.argv` is assigned and am worried that this is not sufficiently robust (e.g. `sys` must be imported each time).

arosenbe: Works like a dream, since the logger scripts import sys. Let me know if you want to see more. 
screen shot 2018-01-12 at 1 45 57 pm

stanfordquan: I'm good, go ahead. 🏒

arosenbe: Can you approve the review please?

Issue #147, opened 2018-05-16
Logging breaks if a folder has space in its name


stanfordquan:
This is because the bash command breaks when you pass a folder with space in its name as we don't wrap it in quotes.

```bash
rclone copy gsbox:gslab_data/raw/divergence/gentzkow_aer_2007 build/external/gentzkow_2007 > build/external/gentzkow_2007/raw data/sconscript_export_gentzkow.log 2>&1
```

It's fairly standard to force a user to not have space characters in file and folder names, so I'm fine with not addressing this issue and instead making a note in template saying "we don't do this."

arosenbe:
Thanks @stanfordquan! Fix above.

Issue #146, opened 2018-05-16
make_list_if_string-related bug


stanfordquan:
@arosenbe I noticed that [this line](https://github.com/gslab-econ/gslab_python/blob/master/gslab_scons/builders/gslab_builder.py#L43) breaks if the input is `[]` instead of `['']` or `''`. Please let me know if this is not a quick fix as expected. Thanks!

stanfordquan:
It's a pretty straightforward "IndexError" because an empty list does not get acted on in `make_list_if_string()` so it wouldn't pass indexing at `[0]`.

```
scons: Reading SConscript files ...
IndexError: list index out of range:
  File "/Users/lequan/divergence/analysis/SConstruct", line 49:
    SConscript('%s/SConscript' % source['external'])
  File "/anaconda/lib/python2.7/site-packages/scons-2.5.1/SCons/Script/SConscript.py", line 604:
    return method(*args, **kw)
  File "/anaconda/lib/python2.7/site-packages/scons-2.5.1/SCons/Script/SConscript.py", line 541:
    return _SConscript(self.fs, *files, **subst_kw)
  File "/anaconda/lib/python2.7/site-packages/scons-2.5.1/SCons/Script/SConscript.py", line 250:
    exec _file_ in call_stack[-1].globals
  File "/Users/lequan/divergence/analysis/source/external/SConscript", line 68:
    main()
  File "/Users/lequan/divergence/analysis/source/external/SConscript", line 8:
    export_hendren()
  File "/Users/lequan/divergence/analysis/source/external/SConscript", line 40:
    env=env)
  File "//anaconda/lib/python2.7/site-packages/gslab_scons/builders/build_anything.py", line 44:
    builder = AnythingBuilder(target, source, action, env, warning, **builder_attributes)
  File "//anaconda/lib/python2.7/site-packages/gslab_scons/builders/build_anything.py", line 62:
    super(AnythingBuilder, self).__init__(target, source, env, name = name)
  File "//anaconda/lib/python2.7/site-packages/gslab_scons/builders/gslab_builder.py", line 43:
    self.source_file      = str(misc.make_list_if_string(source)[0])
```

arosenbe:
Thanks @stanfordquan! Fix above.

PR #148, opened 2018-05-17
Pull request for #146: make_list_if_string-related bug

arosenbe:
@stanfordquan, can you review my fix in #146. I added a check if the `source` argument evaluates to `False`. If so, then we set the source to `''`.  Otherwise, we continue as we did before. 

I also added quoting around the source file to handle the type of issue you raised in #147.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# GSLab Python Library Collection 4.1.0
+# GSLab Python Library Collection 4.1.1
 
 Overview
 --------
--- a/gslab_scons/builders/build_anything.py
+++ b/gslab_scons/builders/build_anything.py
@@ -39,15 +39,17 @@ def build_anything(target, source, action, env, warning = True, **kw):
     builder_attributes = {
         'name': 'Anything Builder'
     }
+    target = [t for t in misc.make_list_if_string(target) if t]
+    source = [s for s in misc.make_list_if_string(source) if s]
     for k, v in kw.items():
         env[k] = v
     builder = AnythingBuilder(target, source, action, env, warning, **builder_attributes)
     bkw = {
-        'action': builder.execute_system_call,
+        'action': builder.build_anything,
         'target_factory' : env.fs.Entry,
         'source_factory':  env.fs.Entry,
     }
-    bld = SCons.Builder.Builder(**bkw) 
+    bld = SCons.Builder.Builder(**bkw)
     return bld(env, target, source)
 
 class AnythingBuilder(GSLabBuilder):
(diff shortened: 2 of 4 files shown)


stanfordquan:
@arosenbe I just found this issue so I'm documenting it here. I haven't checked it out on this PR yet, but FYI.

In `build_anything`, `source = ['']` defaults `source` to the folder the SConscript is in.

```
analysis (master)$ python run.py
scons: done reading SConscript files.
scons: Building targets ...
execute_system_call(["build/prepare_data/data.txt"], ["source/prepare_data"])
scons: `release' is up to date.
scons: done building targets.
analysis (master)$ git diff source/
diff --git a/analysis/source/prepare_data/SConscript b/analysis/source/prepare_data/SConscript
index 88852c6..0638341 100644
--- a/analysis/source/prepare_data/SConscript
+++ b/analysis/source/prepare_data/SConscript
@@ -4,9 +4,9 @@ import gslab_scons as gs
 build  = env['CONFIG']['global']['build']['prepare_data']
 source = env['CONFIG']['global']['source']['prepare_data']

-env.BuildPython(target  = ['#%s/data.txt'       % build],
-                source  = ['#%s/create_data.py' % source],
-                log_ext =  'create_data')
+# env.BuildPython(target  = ['#%s/data.txt'       % build],
+#                 source  = ['#%s/create_data.py' % source],
+#                 log_ext =  'create_data')

 # env.BuildStata(target  = ['#%s/data.txt'       % build],
 #                source  = ['#%s/create_data.do' % source],
@@ -21,10 +21,10 @@ env.BuildPython(target  = ['#%s/data.txt'       % build],
 #                 log_ext =  'create_data')

 # # Invalid syntax for Windows.
-# gs.build_anything(target  = ['#%s/data.txt' % build],
-#                   source  = [''],
-#                   action  = 'sh source/prepare_data/create_data.sh',
-#                   log_ext = 'create_data',
-#                   env     = env)
+gs.build_anything(target  = ['#%s/data.txt' % build],
+                  source  = [''],
+                  action  = 'sh source/prepare_data/create_data.sh',
+                  log_ext = 'create_data',
+                  env     = env)

 env.Install('#release/lg', '#%s/data.txt' % build)
```

arosenbe:
Thanks @stanfordquan! I think the commit above takes care of this. The solution is to remove all the stuff that evaluates to False in the source and target lists before creating the Builder object. I tested this in template, but it'd be very nice to have your confirmation that it works in practice. 

I also tackled a similar issue where we were printing `execute_system_call` in place of `build_anything`. This was as simple as renaming a method.

arosenbe:
@stanfordquan, 

1. I didn't mean to introduce wrapping around the CL args (stray code). I've removed that. As I wrote [above](https://github.com/gslab-econ/gslab_python/pull/148#issuecomment-389751983), I tested in template, and didn't find any errors. But it'd be great to have confirmation in your environment where you noticed the bug. 
2. There's no need to wrap target_dir because it isn't directly used when we execute the system call (I think, please confirm). Like I said above, I implemented quote-wrapping around the log file in #147 [(code)](https://github.com/gslab-econ/gslab_python/blob/issue147_targetspace/gslab_scons/builders/gslab_builder.py#L78). Did you mean something else?

stanfordquan:
We've agreed to implement the quote-wrapping around system calls rather than objects stored within the builder. That means I can pull this in with above changes, block #149 , and implement the quote-wrapping in #147 .

PR #149, opened 2018-05-17
Pull request for #147: Logging breaks if a folder has space in its name

arosenbe:
Thanks @stanfordquan! Can you review my work from #147. The issue was that spaces in the log file weren't correctly formatted for execution on the shell. I fixed this by wrapping the log file in quotes. 

This also required a small change in how we find the log file for time-stamping.


stanfordquan:
See discussion in #148

PR #152, opened 2018-05-17
Pull request for #147: Add os.path.normpath to log and source path calls

stanfordquan:
@arosenbe , as discussed, I added `os.path.normpath` to when paths are called in builders. Thanks! 👩‍🎨

### DIFF
--- a/gslab_scons/builders/build_anything.py
+++ b/gslab_scons/builders/build_anything.py
@@ -48,8 +48,8 @@ def build_anything(target, source, action, env, warning = True, **kw):
     builder = AnythingBuilder(target, source, action, local_env, warning, **builder_attributes)
     bkw = {
         'action': builder.build_anything,
-        'target_factory' : local_env.env.fs.Entry,
-        'source_factory':  local_env.env.fs.Entry,
+        'target_factory' : local_env.fs.Entry,
+        'source_factory':  local_env.fs.Entry,
     }
     bld = SCons.Builder.Builder(**bkw)
     return bld(local_env, target, source)
@@ -93,7 +93,7 @@ class AnythingBuilder(GSLabBuilder):
     def add_call_args(self):
         '''
         '''
-        args = '%s > %s 2>&1' % (self.action, self.log_file)
+        args = '%s > %s 2>&1' % (self.action, os.path.normpath(self.log_file))
         self.call_args = args
         return None
 
--- a/gslab_scons/builders/build_latex.py
+++ b/gslab_scons/builders/build_latex.py
@@ -36,6 +36,6 @@ class LatexBuilder(GSLabBuilder):
         '''
         '''
         target_name = os.path.splitext(self.target[0])[0]
-        args = '%s %s %s > %s' % (self.cl_arg, target_name, self.source_file, self.log_file)
+        args = '%s %s %s > %s' % (self.cl_arg, target_name, os.path.normpath(self.source_file), os.path.normpath(self.log_file))
         self.call_args = args
         return None
--- a/gslab_scons/builders/build_lyx.py
+++ b/gslab_scons/builders/build_lyx.py
@@ -35,7 +35,7 @@ class LyxBuilder(GSLabBuilder):
     def add_call_args(self):
         '''
         '''
-        args = '%s %s > %s' % (self.cl_arg, self.source_file, self.log_file)
+        args = '%s %s > %s' % (self.cl_arg, os.path.normpath(self.source_file), os.path.normpath(self.log_file))
         self.call_args = args
         return None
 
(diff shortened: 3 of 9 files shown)


gslab_scons/builders/build_latex.py
```diff
@@ -36,6 +36,6 @@ def add_call_args(self):
         '''
         '''
         target_name = os.path.splitext(self.target[0])[0]
-        args = '%s %s %s > %s' % (self.cl_arg, target_name, self.source_file, self.log_file)
+        args = '%s %s %s > %s' % (self.cl_arg, target_name, os.path.normpath(self.source_file), os.path.normpath(self.log_file))
```
arosenbe: Do we need to norm the `target_name`?

gslab_scons/builders/build_lyx.py
```diff
@@ -45,6 +45,6 @@ def do_call(self):
         '''
         super(LyxBuilder, self).do_call()
         new_pdf = os.path.splitext(self.source_file)[0] + '.pdf'
-        new_pdf_path = '%s/%s' % (self.target_dir, os.path.basename(new_pdf))
+        new_pdf_path = os.path.normpath('%s/%s' % (self.target_dir, os.path.basename(new_pdf)))
         shutil.move(new_pdf, new_pdf_path)
         return None
```
arosenbe: Would you mind adding a carriage return?

gslab_scons/builders/build_tables.py
```diff
@@ -52,8 +52,8 @@ def do_call(self):
         '''
         '''
         output = tablefill(input    = self.input_string, 
-                           template = self.source_file, 
-                           output   = self.target_file)
+                           template = os.path.normpath(self.source_file), 
+                           output   = os.path.normpath(self.target_file))
```
arosenbe: Are you sure we need to norm these paths?

gslab_scons/builders/build_anything.py
```diff
@@ -48,8 +48,8 @@ def build_anything(target, source, action, env, warning = True, **kw):
     builder = AnythingBuilder(target, source, action, local_env, warning, **builder_attributes)
     bkw = {
         'action': builder.build_anything,
-        'target_factory' : local_env.env.fs.Entry,
-        'source_factory':  local_env.env.fs.Entry,
+        'target_factory' : local_env.fs.Entry,
+        'source_factory':  local_env.fs.Entry,
```
arosenbe: I assume there's a reason for this change.

stanfordquan:
oops

gslab_scons/builders/build_latex.py
stanfordquan: That's just a "name" so I think we're fine?

gslab_scons/builders/build_tables.py
stanfordquan: for consistency + better safe than sorry

gslab_scons/builders/build_anything.py
stanfordquan: I don't think the original was ever correct. `local_env` is in itself an `env` object, so we can't call another `env` on top.

stanfordquan:
```python
AttributeError: 'SConsEnvironment' object has no attribute 'env':
  File "/Users/lequan/template/analysis/SConstruct", line 50:
    SConscript('source/prepare_data/SConscript')
  File "/Users/lequan/template/config/scons/scons-local-3.0.1/SCons/Script/SConscript.py", line 614:
    return method(*args, **kw)
  File "/Users/lequan/template/config/scons/scons-local-3.0.1/SCons/Script/SConscript.py", line 551:
    return _SConscript(self.fs, *files, **subst_kw)
  File "/Users/lequan/template/config/scons/scons-local-3.0.1/SCons/Script/SConscript.py", line 256:
    call_stack[-1].globals)
  File "/Users/lequan/template/analysis/source/prepare_data/SConscript", line 28:
    env     = env)
  File "/Users/lequan/anaconda/lib/python2.7/site-packages/gslab_scons/builders/build_anything.py", line 51:
    'target_factory' : local_env.env.fs.Entry,
```

gslab_scons/builders/build_tables.py
stanfordquan: I checked the code of [tablefill.py](https://github.com/gslab-econ/gslab_python/blob/master/gslab_fill/tablefill.py) and the only time the paths are used is to pass to opening the file (no manual string manipulation), so there should be no worry about platforms for this one.

arosenbe:
Approved 🎆 🍾

stanfordquan:
⛱ ⛱ ⛱ ⛱ ⛱ ⛱

Issue #154, opened 2018-07-06
#154 Installation is unsuccessful every other time


DavidRitzwoller:
Opening an issue for the following error:
![screen shot 2018-07-06 at 2 13 51 pm](https://user-images.githubusercontent.com/35339599/42400589-ff9defcc-8126-11e8-8563-91c97eebb84f.png)

The bug may be localized to anaconda users.

DavidRitzwoller:
@arosenbe confirms that the issue is not confined to anaconda users. I am investigating whether it is associated with the [deletion of old versions of GSLab_Tools. ](https://github.com/gslab-econ/gslab_python/issues/122)

stanfordquan:
Pulled into dev branch on #155 🦉

PR #155, opened 2018-07-06
Pull request for #154: Installation is unsuccessful every other time

DavidRitzwoller:
In conjunction with the changes I've committed [here](https://github.com/gslab-econ/template/pull/107), the pip 10 issues are fixed. The issue was that the requirements.txt installation uses egg, whereas pip uses wheel. Now gslab_python is installed with the wheel, removing the need to manually delete old versions of GSLab_tools implemented [here](https://github.com/gslab-econ/gslab_python/issues/122).


DavidRitzwoller:
@stanfordquan I made the following changes to gslab_python
- I updated the README to instruct the reader how to install from the root of the directory i.e use `pip install .` rather than `python setup.py install clean`. I also added language clarifying why `python setup.py` install should not be called and when to use `python setup.py clean`.
- As mentioned [here](https://github.com/gslab-econ/gslab_python/pull/155#issue-199868025) I removed the code in setup.py which manually cleans `site-packages` of all gslab packages at the beginning of each run.

README.md
```diff
@@ -42,10 +42,19 @@ say, a `sudo` command.
...
+```
+
+We do not reccommend that these packages be installed by executing
+```bash
+python setup.py install
+```
+This method of installation uses egg files rather than Wheels, which can cause conflicts with previous versions of `gslab_tools`. If this method of installation is executed, some files need to be removed from the directory with a `clean` argument. `clean` removes `/build`,`/dist`, and `GSLab_Tools.egg-info`, which are built upon installation. This argument can be called by executing 
+
+```bash
+python setup.py clean
```
stanfordquan: I think this has to be 

```
python setup.py install clean
```

right?

DavidRitzwoller: No, `python setup.py clean` just cleans the repository, including the install argument also reinstalls the packages. setup.py accepts the arguments install, test, and clean.

stanfordquan: Ah, got it. Thanks!

stanfordquan:
Thanks @DavidRitzwoller. Can you please merge this into dev4.1.1? We can pull this in and I'll pull in the template branch once we pull in 4.1.1.

Issue #176, opened 2024-06-19
SaveData python future warning: .fillna()


zhizhongpu:
Running SaveData.py results in:
> SaveData.py:48: FutureWarning: Downcasting object dtype arrays on .fillna, .ffill, .bfill is deprecated and will change in a future version. Call result.infer_objects(copy=False) instead. To opt-in to the future behavior, set `pd.set_option('future.no_silent_downcasting', True)`
>   var_stats = df.describe(include='all').transpose().fillna('')

liaochris:
In this issue, we resolved warnings about downcasting object types when using `fillna()` by updating [`SaveData.py`](https://github.com/gslab-econ/gslab_python/blob/41c620b3274b012b4b1a28995a5242e1d3cf887b/gslab_misc/SaveData/SaveData.py).

- Permanent version of [issue subfolder](https://github.com/gslab-econ/gslab_python/tree/98a0d73c5c10ce377b736daf7448756d72bcf771/issue176_fillna_warning) used for testing
- Latest version of issue commit: https://github.com/gslab-econ/gslab_python/commit/aad359b4ceb6d83d2f82444b903fd4cc28bec0f8

PR #186, opened 2024-08-01
PR for #176: Updated SaveData.py to remove warnings about future deprecation

liaochris:
In #176, following the generic solution in https://stackoverflow.com/a/78066237, we updated the code to remove the warning about future deprecation of features. 
- In the temporary folder `issue176_fillna_warning` there is a notebook where you can run SaveData to ensure that the warning no longer appears. 
- I will delete the temporary folder `issue176_fillna_warning` once you finish reviewing, before I merge. This temporary folder is necessary because without it, there's no way of testing whether the warning shows up (it's not something we can test for using unit tests)


liaochris:
Updated from master in https://github.com/gslab-econ/gslab_python/commit/98a0d73c5c10ce377b736daf7448756d72bcf771

Issue #177, opened 2024-06-27
SaveData does not take pathlib paths


zhizhongpu:
When `out_file` or `log_file` take in Path objects, SaveData runs into error:

```
TypeError: expected string or bytes-like object, got 'PosixPath'
```

The temporary fix is to first transform these arguments using `str()` before passing them into the function. 

Here we explore the possibility to allow SaveData to take `PosixPath` objects directly.

liaochris:
Updated from master in https://github.com/gslab-econ/gslab_python/commit/e3d66eb5d78cd077b1c7cad056c51204c6d9bdd0

liaochris:
In this issue we updated [`SaveData.py`](https://github.com/gslab-econ/gslab_python/blob/e3d66eb5d78cd077b1c7cad056c51204c6d9bdd0/gslab_misc/SaveData/SaveData.py) to take in Path objects as inputs for `out_file` and `log_file`. 

latest version of issue branch: https://github.com/gslab-econ/gslab_python/commit/0cb01cd51fa5da74306da0c95c45c80fec49f565

PR #187, opened 2024-08-01
PR for #177: Updated SaveData to take in pathlib paths

liaochris:
In this issue, I updated `SaveData.py` to take in pathlib path objects (which are of type `PosixPath`) when outputting the file. The error arose because we were trying to verify the file extension using regex, which only accepts strings as input. To address this, I added a separate case that gets the extension using `PosixPath` object methods if the provided path is of type `PosixPath`. I also added a test case that provides as input to the `SaveData` function a `PosixPath`. 

Changes to review are in [`test_main.py`](https://github.com/gslab-econ/gslab_python/blob/0472b924b84e515793f17e0ef2d0add97300c1bf/gslab_misc/SaveData/tests/test_main.py)


liaochris:
summary comment: https://github.com/gslab-econ/gslab_python/issues/177#issuecomment-2319132504

